PR #17, opened 2019-06-08
Use arenas and TokenReference instead of Cow<Token>

### LINKED ISSUE #16: Change Cow<'_, Token<'_>> to Arc<RwLock<Token<'_>>> or something similar
A requirement for mutability is for tokens to be references instead of clones of one another. This seems like the most reasonable way to do it.

Kampfkarren:
Tokenization may have gotten slower, but I'm not sure by how much. ParserState is no longer a simple Copy, but now contains an Arc that needs to be cloned.

Closes #16, and opens up the door for mutability.

TODO:
- [x] Implement PartialOrd/Ord for Position, and then Token
- [x] Change `Ast::iter_tokens` to sort by position, as Arena iteration is undefined
- [x] Figure out how `VisitMut` should be implemented for `TokenReference`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changed
 - Fields of `Token` and `Position` have been made private with public accessors
+- Added `TokenKind` to get the kind of a token without any additional data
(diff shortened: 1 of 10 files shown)


Kampfkarren:
`cargo fmt` made some really ugly stuff in `ast.rs`, like multi-line lifetimes. Not worth caring about, but disappointing.

Kampfkarren:
Hope this works.

PR #20, opened 2019-07-08
New Node trait - Closes #6

### LINKED ISSUE #6: Get full range of node
As in the full range of `local x, y = 1, 2` would start at "local" and end at "2".

Kampfkarren:
Will need to also add some new properties to keep track of tokens (such as the `local` in a `LocalAssignment`)

TODO:
- [ ] Rerun tests
~~- [ ] Add methods for all the new properties~~
~~- [ ] Reconsider start_position/end_position names, it conflicts with Token.~~


Kampfkarren:
TODO: Remove EXPANDED.rs, I was debugging.

Kampfkarren:
Should probably rename `start_position` and `end_position`, it conflicts with `Token`.

Kampfkarren:
Might not totally fill every token and whatnot here, just a few. In the future I want something like syn has with `Punctuated`. https://oschwald.github.io/maxminddb-rust/syn/punctuated/struct.Punctuated.html

PR #30, opened 2019-08-13
LastStmt::Return to use its own Return struct

Kampfkarren:
Inspired by syn.


Kampfkarren:
syn doesn't eliminate them entirely, so I'll keep a few that make sense.

Kampfkarren:
Actually, just this didn't make sense, I guess.

PR #32, opened 2019-08-13
More properties

Kampfkarren:


Kampfkarren:
Next after this is getters.

PR #51, opened 2019-12-15
Refactor some tokenize functions to use nom

tversteeg:
The small unit tests in the file are succeeding but the `multi-line-comments-3` test in `tests/` is broken for some reason.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ lazy_static = "1.3"
 paste = "0.1"
 regex = "1.1"
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
-nom = "5.0.1"
+nom = "5.0"
 
 [dev-dependencies]
 criterion = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ lazy_static = "1.3"
 paste = "0.1"
 regex = "1.1"
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
+nom = "5.0.1"
 
 [dev-dependencies]
 criterion = "0.2"
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -2,6 +2,13 @@ use crate::visitors::{Visit, VisitMut, Visitor, VisitorMut};
 use atomic_refcell::AtomicRefCell;
 use generational_arena::{Arena, Index};
 use lazy_static::lazy_static;
+use nom::{
+    bytes::complete::take_while,
+    character::{complete::alpha1, is_alphanumeric},
+    combinator::recognize,
+    sequence::pair,
+    IResult,
+};
 use regex::{self, Regex};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
@@ -666,10 +673,26 @@ fn advance_number(code: &str) -> Advancement {
     })
 }
 
+#[inline]
+fn parse_identifier(code: &str) -> IResult<&str, &str> {
+    recognize(pair(
+        // Identifiers must start with at least 1 alphabetic character
+        alpha1,
+        // And then they must be followed by 0 or more alphanumeric (or '_') characters
+        take_while(|x| is_alphanumeric(x as u8) || x == '_'),
+    ))(code)
+}
+
 fn advance_identifier(code: &str) -> Advancement {
-    advance_regex!(code, PATTERN_IDENTIFIER, Identifier(find) {
-        identifier: Cow::from(find.as_str()),
-    })
+    match parse_identifier(code) {
+        Ok((_, identifier)) => Ok(Some(TokenAdvancement {
+            advance: identifier.chars().count(),
+            token_type: TokenType::Identifier {
+                identifier: Cow::from(identifier),
+            },
+        })),
+        Err(_) => Ok(None),
+    }
 }
 
 fn advance_quote(code: &str) -> Advancement {
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -1,10 +1,11 @@
 use crate::visitors::{Visit, VisitMut, Visitor, VisitorMut};
 use atomic_refcell::AtomicRefCell;
+use full_moon_derive::symbols;
 use generational_arena::{Arena, Index};
 use lazy_static::lazy_static;
 use nom::{
     branch::alt,
-    bytes::complete::{take_while, take_while1},
+    bytes::complete::{tag, take_while, take_while1},
     character::complete::{line_ending, space1},
     combinator::{opt, recognize},
     sequence::pair,
@@ -23,53 +24,6 @@ use std::{
         Arc,
     },
 };
-
-macro_rules! symbols {
-    ($($ident:ident => $string:tt,)+) => {
-        /// A literal symbol, used for both words important to syntax (like while) and operators (like +)
-        #[derive(Clone, Copy, Debug, Eq, PartialEq)]
-        #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
-        pub enum Symbol {
-            $(
-                #[cfg_attr(feature = "serde", serde(rename = $string))]
-                #[allow(missing_docs)]
-                $ident,
-            )+
-        }
-
-        impl<'a> fmt::Display for Symbol {
-            fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-                match *self {
-                    $(Symbol::$ident => $string,)+
-                }
-                .fmt(formatter)
-            }
-        }
-
-        impl FromStr for Symbol {
-            type Err = ();
-
-            fn from_str(string: &str) -> Result<Self, Self::Err> {
-                Ok(match string {
-                    $($string => Symbol::$ident,)+
-                    _ => return Err(()),
-                })
-            }
-        }
-
-        lazy_static! {
-            static ref PATTERN_SYMBOL: Regex = Regex::new(
-                &vec![$($string,)+]
-                    .iter()
-                    .map(|x| {
-                        regex::escape(&x.to_string())
-                    })
-                    .collect::<Vec<_>>()
-                    .join("|")
-            ).unwrap();
-        }
-    };
-}
 
 symbols!(
     And => "and",
@@ -764,27 +718,15 @@ fn advance_quote(code: &str) -> Advancement {
 }
 
 fn advance_symbol(code: &str) -> Advancement {
-    if code.chars().next().unwrap().is_ascii_alphanumeric() {
-        let identifier = match parse_identifier(code) {
-            Ok((_, identifier)) => identifier,
-            Err(_) => panic!("Parsing identifier failed"),
-        };
+    match parse_symbol(code) {
+        Ok((_, string)) => Ok(Some(TokenAdvancement {
+            advance: string.chars().count(),
+            token_type: TokenType::Symbol {
+                symbol: Symbol::from_str(string).unwrap(),
+            },
+        })),
 
-        let expected_len = identifier.len();
-
-        advance_regex!(&code[0..expected_len], PATTERN_SYMBOL, Symbol(find) {
-            symbol: {
-                if find.end() - find.start() == expected_len {
-                    Symbol::from_str(find.as_str()).unwrap()
-                } else {
-                    return Ok(None)
-                }
-            }
-        })
-    } else {
-        advance_regex!(code, PATTERN_SYMBOL, Symbol(find) {
-            symbol: Symbol::from_str(find.as_str()).unwrap(),
-        })
+        Err(_) => Ok(None),
     }
 }
 
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -592,19 +592,19 @@ fn parse_multi_line_comment_start(code: &str) -> IResult<&str, &str> {
 #[inline]
 fn parse_multi_line_comment_body<'a>(
     code: &'a str,
-    equality_symbol: &'a str,
+    block_count: &'a str,
 ) -> IResult<&'a str, &'a str> {
     recognize(many_till(
         anychar,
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -25,6 +25,7 @@ lazy_static = "1.3"
 paste = "0.1"
 regex = "1.1"
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
+nom = "5.0.1"
```
Kampfkarren: Probably don't want to pin a specific version like this, assuming they follow semver that is.

Kampfkarren:
I'll take a look at why the test is failing, thanks for the PR!

src/tokenizer.rs
```diff
@@ -666,10 +673,26 @@ fn advance_number(code: &str) -> Advancement {
...
+    recognize(pair(
+        // Identifiers must start with at least 1 alphabetic character
+        alpha1,
+        // And then they must be followed by 0 or more alphanumeric (or '_') characters
+        take_while(|x| is_alphanumeric(x as u8) || x == '_'),
+    ))(code)
+}
+
 fn advance_identifier(code: &str) -> Advancement {
-    advance_regex!(code, PATTERN_IDENTIFIER, Identifier(find) {
```
Kampfkarren: Remove the PATTERN_IDENTIFIER constant if we don't need it anymore

Kampfkarren:
For me, it's numeric-for-loop that errors, not multi-line-comments. I wonder if I somehow broke something with your version when I released https://github.com/Kampfkarren/full-moon/commit/77514782a47d974ca534feb0f24dc06ae2700624 and its multi-line optimizations? Does the test fail on Linux?

src/tokenizer.rs
```diff
@@ -666,10 +673,26 @@ fn advance_number(code: &str) -> Advancement {
     })
 }
 
+#[inline]
+fn parse_identifier(code: &str) -> IResult<&str, &str> {
+    recognize(pair(
+        // Identifiers must start with at least 1 alphabetic character
+        alpha1,
+        // And then they must be followed by 0 or more alphanumeric (or '_') characters
+        take_while(|x| is_alphanumeric(x as u8) || x == '_'),
```
Kampfkarren: This could be replaced with `x.is_ascii_alphanumeric()` -- https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric

Kampfkarren:
I figured out the issue--if a variable starts with `_`, your nom combinator doesn't catch it since it only allows starting with an alphabetical character!

tversteeg:
Thanks a lot that fixed it!

Kampfkarren:
Sweet, I'll see if the benchmarks improved.

Kampfkarren:
Hmm...`cargo bench` reports that this parses date *slower* than the Regex :(

I wonder if replacing this with just direct code would have an improvement?

tversteeg:
For me it improved with 6%..

Kampfkarren:
That's strange, the inconsistent results might suggest this isn't that much of a bottleneck, I'm going to do more testing on this for now.

tversteeg:
Yeah, but of course it's only a small part at the moment, I'll see if I can replace more with nom.

Kampfkarren:
Just whipped this up, it passes tests, can you see if it beats nom's performance for you?

```rs
fn advance_identifier(code: &str) -> Advancement {
    let mut chars = code.chars();
    let initial = chars.next().unwrap();

    if initial == '_' || initial.is_ascii_alphabetic() {
        let end = chars
            .take_while(|c| *c == '_' || c.is_ascii_alphanumeric())
            .count()
            + 1;

        Ok(Some(TokenAdvancement {
            advance: end,
            token_type: TokenType::Identifier {
                identifier: Cow::Borrowed(&code[0..end]),
            },
        }))
    } else {
        Ok(None)
    }
}
```

tversteeg:
For me it's actually worse now with your change, it regressed 4.5% for the AST test.

Kampfkarren:
Weird, keep at it with nom then!

tversteeg:
I'm currently trying to replace the `PATTERN_SYMBOL` in `symbols!` with the following:

```rust
impl Symbol {
    pub fn find(code: &str) -> IResult<&str, Option<Self>> {
        alt(($(tag($string),)+))(code)
    }
}
```

But this doesn't work because `alt` only accepts [21 arguments](https://github.com/Geal/nom/issues/994) (and there are a lot more keywords than that). There's a possibility to split the `alt` call into multiple calls with smaller groups. The problem with this is that my Rust-macro-foo isn't good enough to do this procedurally. Any suggestions?

Kampfkarren:
> Isn't another problem the amount of allocations that have to be done because of Cow though? Maybe it would be better to have a &'a str reference pointing to the original source code &str so no extra allocations need to be done.

Didn't see this comment, but I'm pretty sure we need the `Cow` because in full-moon, ASTs can be mutated to add new (owned) tokens and such.

As for your `alt` woes, it's possible to just shove it into `full_moon_derive` (our source for proc macros), but that doesn't sound like a great idea to me. You could manually split them yourself, so like `symbols!` looks like:

```rust
symbols!(
	{
    	And => "and",
    	Break => "break",
    	Do => "do",
    	ElseIf => "elseif",
    	Else => "else",
    	End => "end",
    	False => "false",
    	For => "for",
    	Function => "function",
    	If => "if",
    	In => "in",
    	Local => "local",
    	Nil => "nil",
    	Not => "not",
    	Or => "or",
    	Repeat => "repeat",
    	Return => "return",
    	Then => "then",
    	True => "true",
    	Until => "until",
    	While => "while",
	},

	{
		Caret => "^",
		Colon => ":",
		Comma => ",",
		Ellipse => "...",
		TwoDots => "..",
		Dot => ".",
		TwoEqual => "==",
		Equal => "=",
		GreaterThanEqual => ">=",
		GreaterThan => ">",
		Hash => "#",
		LeftBrace => "{",
		LeftBracket => "[",
		LeftParen => "(",
		LessThanEqual => "<=",
		LessThan => "<",
		Minus => "-",
		Percent => "%",
		Plus => "+",
		RightBrace => "}",
		RightBracket => "]",
	},

	{
		RightParen => ")",
		Semicolon => ";",
		Slash => "/",
		Star => "*",
		TildeEqual => "~=",
	},
);
```

It's not great, though.

tversteeg:
> As for your `alt` woes, it's possible to just shove it into `full_moon_derive` (our source for proc macros), but that doesn't sound like a great idea to me. You could manually split them yourself, so like `symbols!` looks like:
> 
> ```rust
> symbols!(
> 	{
>     	And => "and",
>     	Break => "break",
>     	Do => "do",
>     	ElseIf => "elseif",
>     	Else => "else",
>     	End => "end",
>     	False => "false",
>     	For => "for",
>     	Function => "function",
>     	If => "if",
>     	In => "in",
>     	Local => "local",
>     	Nil => "nil",
>     	Not => "not",
>     	Or => "or",
>     	Repeat => "repeat",
>     	Return => "return",
>     	Then => "then",
>     	True => "true",
>     	Until => "until",
>     	While => "while",
> 	},
> 
> 	{
> 		Caret => "^",
> 		Colon => ":",
> 		Comma => ",",
> 		Ellipse => "...",
> 		TwoDots => "..",
> 		Dot => ".",
> 		TwoEqual => "==",
> 		Equal => "=",
> 		GreaterThanEqual => ">=",
> 		GreaterThan => ">",
> 		Hash => "#",
> 		LeftBrace => "{",
> 		LeftBracket => "[",
> 		LeftParen => "(",
> 		LessThanEqual => "<=",
> 		LessThan => "<",
> 		Minus => "-",
> 		Percent => "%",
> 		Plus => "+",
> 		RightBrace => "}",
> 		RightBracket => "]",
> 	},
> 
> 	{
> 		RightParen => ")",
> 		Semicolon => ";",
> 		Slash => "/",
> 		Star => "*",
> 		TildeEqual => "~=",
> 	},
> );
> ```
> 
> It's not great, though.

This is exactly what I tried, but I couldn't get the nested macro's to work. Do you know how this can be done?

Kampfkarren:
I'm not sure a great way to explain how to do it, but I whipped up a nested macro as a proof of concept, this calls `add(1, add(2, add(3, 0)))` in the end.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9a63db24b720a6aacd208f708091fdd8

tversteeg:
> I'm not sure a great way to explain how to do it, but I whipped up a nested macro as a proof of concept, this calls `add(1, add(2, add(3, 0)))` in the end.
> 
> [play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9a63db24b720a6aacd208f708091fdd8](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9a63db24b720a6aacd208f708091fdd8)

Thanks for the example but I think it will be a bit inefficient performance-wise to have a separate `alt` for each keyword. Grouping them by groups of 20 keywords would be more ideal, similar to the `symbols!` example you posted above.

Kampfkarren:
I didn't intend for you to do it one by one, I just wanted to show a simple example to help out. 🙂

tversteeg:
Oh haha in that case you can merge it if you're happy with it :)

Kampfkarren:
No no, I meant as in with my example I didn't expect you to do:

`alt(foo, alt(bar, alt(baz)))`

I intended for you to do the thing with multiple groups, but just using the `add` thing as an example for how to do that.

tversteeg:
Yeah but the problem with making it groups, as I tried to explain above, is that I don't know how to parse the brackets in the macro and the contents.

So I'm trying to do something like this, but I can't get it to work:

```rust
macro_rules! extract {
	($($ident:ident => $string:tt,)+) => {
		$($string => Symbol::$ident,)+
	}
}

macro_rules! symbols {
	($($group:block,)+ => {
		...
		
        impl<'a> fmt::Display for Symbol {
            fn fmt(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                match *self {
                    $(extract!($group),)+
                }
                .fmt(formatter)
            }
        }

		...
	}
}

symbols!(
	{
		And => "and",
		...
	}
)
```

Kampfkarren:
I'll see if I can get something working soon.

Kampfkarren:
Ok, this example is a bit more in depth (and ugly), but shows how you would go about doing this. Does this help?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7ffeca997316a937b697973371db6a9e

tversteeg:
It's helping getting me a small step further but now I'm stuck again, maybe you can do this part, this is what I currently have but it's not working at all:

```rust
fn add(x: i32, y: i32, z: i32) -> i32 {
    x + y + z
}

macro_rules! add {
    () => (0);

    (
        {
            $($head:expr,)+
        },

        $({
            $($n:expr,)+
        },)*
    ) => {
        add(
            $($head,)+
            add!($({
                $($n,)+
            },)*),
        )
    }
}

fn main() {
    let x = add!(
        {
            1, 2,
        },

        {
            3, 4,
        },
    );

    assert_eq!(x, 10);
}
```

Kampfkarren:
Can you push what you have to your nom branch so I can work with it?

tversteeg:
I haven't got anything new besides the WIP snippet I posted here, but that's not working at all so do you still want me to push that?

Kampfkarren:
In that case, no. I'll take a look when I can to see if it's easy to turn this into a procedural macro.

Kampfkarren:
Sent a commit through, feel free to look at the changes, and benchmark it when you can :)

tversteeg:
Thanks a lot, this really helps!

Is there a specific reason you group all the identifiers together? When I look at the [BNF](http://wiki.luajit.org/Extended-BNF) it seems to make a lot of sense to group them differently, that would greatly simplify the parsing. That way the (seemingly expensive) length comparison check between an operator and something else doesn't need to be done.

Kampfkarren:
I don't remember why exactly, other than it ends up being a lot easier to work with as an API then just using the same things as identifiers, especially since `visit_identifier` is a thing, you might not want to get stuff like `local` thrown in there.

src/tokenizer.rs
```diff
@@ -580,44 +579,55 @@ macro_rules! advance_regex {
...
 
-            let comment = &code[beginning.end()..end_find.start()];
+#[inline]
+fn parse_multi_line_comment_body<'a>(
+    code: &'a str,
+    equality_symbol: &'a str,
+) -> IResult<&'a str, &'a str> {
+    recognize(many_till(
+        anychar,
+        recognize(tuple((tag("]"), tag(equality_symbol), tag("]")))),
```
tversteeg: I tried to find a way here to return the comment without the ending "]]" and without extra allocations but I couldn't find anything.

src/tokenizer.rs
```diff
@@ -620,44 +579,55 @@ macro_rules! advance_regex {
...
+#[inline]
+fn parse_multi_line_comment_start(code: &str) -> IResult<&str, &str> {
+    delimited(tag("--["), take_while(|x: char| x == '='), tag("["))(code)
+}
 
-            let comment = &code[beginning.end()..end_find.start()];
+#[inline]
+fn parse_multi_line_comment_body<'a>(
+    code: &'a str,
+    equality_symbol: &'a str,
```
Kampfkarren: Not a great name, it's not checking the equality of anything, really. Last name was block_count, which isn't great but describes it better.

src/tokenizer.rs
Kampfkarren: `.map(|comment| &comment[0..comment.len() - block_count])`?

tversteeg: That's basically what I do in the block below, so I guess I'll keep it like this.

tversteeg:
I won't be able to work on this PR anymore for at least until February, so if that's a problem for you I would advise to merge it and implement the last parts yourself, which should be quite easy using the code I already pushed as an example.

Kampfkarren:
That's fine, I'll look into how to do that later and merge this. Thanks a lot for your hard work! 😄

Kampfkarren:
Thanks for the help!

PR #61, opened 2020-02-08
Change API and parser to work more like Swift's libsyntax.

### LINKED ISSUE #26: Implement Display for all nodes
Possibly even use that for `print`?

### LINKED ISSUE #60: Consider rewriting Ast to be more like Swift's syntax library
Swift's syntax library makes it *really* easy to create a Display implementation (that could be used by `Ast::print`) and mutate without worrying about the weird arena we have now.

Notably, nodes are entirely self contained.

https://github.com/apple/swift/tree/master/lib/Syntax

Kampfkarren:
Closes #60. Allows for easy mutation, killing the arena, making `ParserState` copy again and non-atomic (read: should be very fast).

Requires a massive codebase rewrite. This'll set the stage for 0.5.0, and hopefully 1.0.0.

TODO:
- [x] Add `with_` methods for everything that needs them
- [x] Add constructor methods to AST nodes to be able to create fresh ones
- [x] Remove `surrounding_ignore_tokens`
- [x] Implement update_positions or some equivalent
- [x] Rename `visit_token` to `visit_token_reference`, or whatever this will be renamed to
- [x] Fix `Visit` and `VisitMut` ordering for Roblox type syntax
- [x] Fix identified regression
- [x] Add all documentation, uncomment docs warning

Won't do:
- [ ] Remove `unsafe` in `peek`
- [ ] Rename TokenReference to something accurate to what it is now, which is an inner token with leading and trailing trivia. Swift uses TokenSyntax

User-facing changelog (added on as time goes on):
- TokenReference has been completely rewritten to now be a representation of an inner token and its leading/trailing trivia. It will eventually be renamed
- Display is now implemented on all nodes (Closes #26)
- `ignore` is now deprecated in favor of `is_trivia`
- `last_stmts` is now deprecated in favor of `last_stmt
- Added `Punctuated::last`
- Fixed a bug where illogical punctuated sequences were created when function bodies had both named parameters and varargs
- Added `TokenReference::with_token`
- Added `Token::new`
- Added `Puncutated::from_iter`
- Added `TokenType::spaces` and `TokenType::tabs`


Kampfkarren:
It's way faster now! 😄 

image

image

tversteeg:
This is looking great!

tversteeg:
Is this currently in a testable state?

Kampfkarren:
Depends on what you mean by testable, I think. It passes the test cases we have, but mutability is still a bit ways off until I add with_ functions to everything. If you are just doing immutable work though, then it should be usable.

tversteeg:
This is impressive, I went from 940 ms for 10 samples on date.lua to 16 ms!

Kampfkarren:
Still working on this. Currently trying to figure out the best way to kill `surrounding_ignore_tokens` and not require the `ast` parameter or the hacky method it uses now. Because nodes have ownership over their tokens now, it's possible. I've evaluated that the best way is probably adding a `tokens` function to `Node`, which even `start_position` and `end_position` can use. From that, there are two ways to implement it:

1. Collect every token owned by the node and all its children and return it. This is the simplest, but I did something like this with the old `*_position` implementations and it led to terrible performance on use. Especially notable since most of the use cases are going to involve the first and/or last nodes.

2. Return a type that implements `Iterator` and `DoubleEndedIterator` that *lazily* evaluates tokens. As in, if a node has a structure like:

```rust
struct Example<'ast> {
    foo_token: Cow<'ast, TokenReference<'ast>>,
    value: Value<'ast>,
}
```

...then the iterator will understand in the context of "an owned token, and another Node that we may have to get the tokens of later". 

Problem is this seems to cause issues with Rust's ownership model that I have to wrap my head around solving. 

I'm not sure if there's any performant alternatives to these two that I am not considering.

tversteeg:
I've found another good test file that doesn't seem to work on my machine:

https://github.com/bluebird75/luaunit/blob/master/luaunit.lua

I get an `unexpected token 'E11'`.

By the way, the performance of this version is phenomenal!

Kampfkarren:
Thanks for the test case, I'll see if I can reproduce it! Have been busy with other work, so haven't had time to work on this PR some more.

Kampfkarren:
@tversteeg Ah, so it's because they use `1E11`, with a capital `e`, something I've never seen before. Looks like we fail for `0XFF` too (capital `0x`). Easy fix, though.

Kampfkarren:
Going to merge this in later when I get the time.

tversteeg:
Yay! Thank a lot.

PR #65, opened 2020-02-27
Make tokenizer less restrictive

Xiphoseer:
This PR includes two small changes to the tokenizer that will make it less restrictive in some cases.

1. Accept number literals that omit a value of `0` before the decimal point. For example `.5` or `.7e2` are now valid number tokens that should be interpreted as if they were `0.5` or `0.7e2` respectively. This notation is accepted by the LUA 5.1 interpreter.

2. Accept `'\r'` as normal whitespace. For example the sequence `"\r\r\n"` currently causes a syntax error while the LUA 5.1 interpreter runs it just fine.
   
   The reason for that is that the `nom` parsers `space1` and `line_ending` don't match the regex used in the LUA grammar. `nom` sees `' '` and `'\t'` as space and `"\r\n"` or `"\n"` as a line break, LUA (according to a comment in this codebase) uses `[^\S\n]` as space (which is `' '`, `'\t'`, `'\r'`, `'\u{b}'`, `\u{c}` if you don't account for unicode whitespace) and `'\n'` as line break.
   
   The changes in this PR are a minimal step in that direction, by adding `'\r'` to be handled like a space and using `tag("\n")` instead of `nom`s `line_ending` for line breaks.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -5,9 +5,9 @@ use generational_arena::{Arena, Index};
 use nom::{
     branch::alt,
     bytes::complete::{tag, take_till, take_while, take_while1},
-    character::complete::{anychar, digit1, line_ending, space1},
+    character::complete::{anychar, digit1, one_of},
     combinator::{opt, recognize},
-    multi::many_till,
+    multi::{many1, many_till},
     sequence::{delimited, pair, preceded, tuple},
     IResult,
 };
@@ -602,10 +602,18 @@ fn parse_hex_number(code: &str) -> IResult<&str, &str> {
     recognize(pair(tag("0x"), take_while1(|c: char| c.is_digit(16))))(code)
 }
 
+fn num_frac(code: &str) -> IResult<&str, &str> {
+    recognize(pair(tag("."), digit1))(code)
+}
+
+fn num_expt(code: &str) -> IResult<&str, &str> {
+    recognize(pair(tag("e"), digit1))(code)
+}
+
 fn parse_basic_number(code: &str) -> IResult<&str, &str> {
     recognize(pair(
-        digit1,
-        pair(opt(pair(tag("."), digit1)), opt(pair(tag("e"), digit1))),
+        alt((num_frac, recognize(pair(digit1, opt(num_frac))))),
+        opt(num_expt),
     ))(code)
 }
 
@@ -758,7 +766,10 @@ fn advance_symbol(code: &str) -> Advancement {
 #[inline]
 fn parse_whitespace(code: &str) -> IResult<&str, &str> {
     // From regex "^[^\S\n]+\n?|\n"
-    alt((recognize(pair(space1, opt(line_ending))), line_ending))(code)
+    alt((
+        recognize(pair(many1(one_of(" \t\r")), opt(tag("\n")))),
+        tag("\n"),
+    ))(code)
 }
 
 // Keep finding whitespace until the line ends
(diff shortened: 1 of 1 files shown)


Kampfkarren:
Thanks for the PR! I'm going to hold off on merging this for now since there's a massive refactor PR going on right now, but this shouldn't cause any conflicts with that. Will make sure your PR works with mine. Thanks!

Xiphoseer:
Great! Thank you for your time and good luck with the refactoring.

Rerumu:
I looked at the code and wanted to point out the PR still doesn't support different cases in `0x`, `0X`, `e` and `E`. Hexadecimal numbers also have an optional exponent that uses `p` instead of `e`. The former issue can probably be solved using `tag_no_case`, but the latter is slightly trickier.

Kampfkarren:
Sorry for the long delay, but the refactor is merged now if you want to take a look at this again.

Kampfkarren:
Stale, let me know if you want to work on this again.

Xiphoseer:
Ah, thanks for the reminder! I somehow missed your last comment and haven't worked on the project I use this for in a while.

I'll look at this and get back to you when I find the time.

I'm also a little disappointed that the performance characteristics of using this library (to process a bunch of files to JSON for syntax highlighting in a browser and possibly some analysis down the line) seems worse than my previous impl which used luaparse.js

I may end up trying whether generating JSON for this AST and rewriting my display code is much faster or whether I'll find a different library that is as complete as this one but less focused on being usable in an IDE and the associated overhead of using Arc

Kampfkarren:
> I'm also a little disappointed that the performance characteristics of using this library (to process a bunch of files to JSON for syntax highlighting in a browser and possibly some analysis down the line) seems worse than my previous impl which used luaparse.js

Have you tried the latest version? Since you made this PR some massive optimizations have been made.

Xiphoseer:
> Have you tried the latest version? Since you made this PR some massive optimizations have been made.

I have not, which is why I'll happily give it a try when I find the time and energy. It was more to hint as to why I may not follow through with using this lib.

My current project is mostly glue code to convince miniserde to produce a luaparse AST from this lib and that really isn't satisfying to write.

Xiphoseer:
I have now tried out the current version (by including the latest commit as a git dependency) and I can confirm that the performance is much better now.

I still had issues with the tokenizer rejecting some random misplaced `\r` characters in the scripts I'm working with. I chose to just fix the few cases on my end, but ideally, the tokenizer would confirm to the LUA spec. If you want, I can extract those changes from this PR and send a new one, just for line endings.

Kampfkarren:
> I still had issues with the tokenizer rejecting some random misplaced \r characters in the scripts I'm working with.

I got word of this too, I hopefully fixed it in #72.

In the meantime, #71 was merged. Can you make sure that it doesn't supersede what you did here?

Xiphoseer:
Yes, I've seen that #71 was merged and it resolves the numbers part of this PR. Sorry that I didn't notice the existing issue for that when I created mine.

As far as I can tell, #72 only touches tests, not the tokenizer code. I've seen one recent PR for CRLF, but the issue I'm facing is single CRs as newlines in some places, not a CRLF.

LUA treats those as simple whitespace and only really cares about the LF to break lines. The tokenizer in this lib just rejects these files with the code in this repo (tested just before my last comment was made)

Kampfkarren:
Definitely file an issue for those, or even better a pull request to fix it. Files with lone `CR`s are rare to my knowledge, so this is the first I'm hearing about it!

PR #66, opened 2020-04-22
Syntax fix

Xeptix:
Two syntax changes included to reflect the new typed lua syntax announced:

1. For function definitions, instead of using a “fat arrow” (=>), now use a colon (:) to delimit the return type
2. For function types, instead of using a “fat arrow” (=>), now use a “thin arrow” (->) to delimit the return type

Modifications shouldn't cause issues however the change was not tested due to complications.

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -38,7 +38,6 @@ symbols!(
     While => "while",
 
     // TODO: This only is valid in Roblox
-    FatArrow => "=>",
     ThinArrow => "->",
     Caret => "^",
     Colon => ":",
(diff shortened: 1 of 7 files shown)


src/tokenizer.rs
```diff
@@ -39,6 +39,7 @@ symbols!(
 
     // TODO: This only is valid in Roblox
     FatArrow => "=>",
+    ThinArrow => "->",
```
Kampfkarren, line 42: This leaves `FatArrow` unused, can you remove that?

Xeptix: Pushed all requested changes.

Kampfkarren:
What test did you run? `cargo test --features roblox` gives me errors on the latest version.

Xeptix:
All 25 tests are now successful on the latest version. I indeed ran the wrong test, all should be well now.

Kampfkarren:
Thanks!

PR #67, opened 2020-05-05
Implement Lua bindings

Kampfkarren:
Will be usable not just as something like jscodeshift, but also for custom Selene lints!

Will be behind an optional feature flag, it's default for now for testing.


Kampfkarren:
Going to close this for now, I'm a bit lost on some finer details on how this would be implemented, but this is definitely a feature I'm going to pick up again.

PR #68, opened 2020-05-12
Fix leading_trivia with CRLF line endings

evaera:
Fixes `leading_trivia` not being populated correctly within `TokenReference`s inside `extract_token_references` due to CRLF line endings

test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out


Kampfkarren:
Love Windows! Thanks!

PR #71, opened 2020-06-29
Improve parsing support for numbers

### LINKED ISSUE #57: New parser doesn't support exponential notation with a negative number
`1e-5` is reported as an error.

### LINKED ISSUE #58: New parser can't parse numbers starting with a point
`local x = .1` is a parse error.

### LINKED ISSUE #63: Allow omitting an integer part of 0 in number literals
The lua5.1 implementation allows numbers literals like `.5` and even `.5e0` to represent a value of `0.5`. The `full-moon` tokenizer currently does not support this and it's not mentioned in the spec that this should work.

I'd like to use `full-moon` for an existing codebase which uses this a lot and would appreciate to know whether you could consider implementing this or accept a pull-request for it.

JohnnyMorganz:
This PR will:
- Add support for negative exponents (eg. `1e-5`) - Closes #57 
- Add support for fractional numbers with no integer part (eg. `.5`) - Closes #58, Closes #63 (Duplicate?)

With tests for both

### DIFF
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -688,14 +688,8 @@ fn parse_no_int_fractional_number(code: &str) -> IResult<&str, &str> {
         opt(digit1),
         pair(
             pair(tag("."), digit1),
-            opt(pair(
-                pair(
-                    tag_no_case("e"),
-                    opt(tag("-"))
-                ),
-                digit1
-            ))
-        )
+            opt(pair(pair(tag_no_case("e"), opt(tag("-"))), digit1)),
+        ),
     ))(code)
 }
 
@@ -704,13 +698,7 @@ fn parse_basic_number(code: &str) -> IResult<&str, &str> {
         digit1,
         pair(
             opt(pair(tag("."), digit1)),
-            opt(pair(
-                pair(
-                    tag_no_case("e"),
-                    opt(tag("-"))
-                ),
-                digit1
-            )),
+            opt(pair(pair(tag_no_case("e"), opt(tag("-"))), digit1)),
         ),
     ))(code)
 }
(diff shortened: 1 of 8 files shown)


full-moon/src/tokenizer.rs
```diff
@@ -683,12 +683,34 @@ fn parse_hex_number(code: &str) -> IResult<&str, &str> {
...
 
+fn parse_no_int_fractional_number(code: &str) -> IResult<&str, &str> {
+    recognize(pair(
+        opt(digit1),
+        pair(
+            pair(tag("."), digit1),
+            opt(pair(
+                pair(
+                    tag_no_case("e"),
+                    opt(tag("-"))
```
Kampfkarren, line 694: There should be a trailing comma here. Make sure you run `cargo fmt`, which will automatically format all your code to the standard.

full-moon/tests/cases/pass/fractional-numbers/source.lua
```diff
@@ -0,0 +1,4 @@
+local num = 0.5
+local num2 = 0.5e5
+local num3 = .5
+local num4 = .5e5
```
Kampfkarren, line 4: Can you include a test for stuff like `1e-5` too? We've had trouble with that.

JohnnyMorganz: That was added as a separate test under the /exponents/ folder, unless you want to merge them all into one big test?

full-moon/src/tokenizer.rs
JohnnyMorganz: I'll remember that for future!

full-moon/tests/cases/pass/fractional-numbers/source.lua
Kampfkarren: I didn't see that, nevermind.

PR #70, opened 2020-06-26
Add compound assignment operators

MicahHinckley:

PR #74, opened 2020-06-30
Add Compound Assignment support

### LINKED ISSUE #69: Add compound assignments to Roblox parser
image

JohnnyMorganz:
Adds support for Compound Assignment using Compound Operators under the roblox feature flag
Based off of #70 (Thanks @MicahHinckley)

Closes #69

### DIFF
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -387,6 +387,10 @@ pub enum Value<'a> {
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 pub enum Stmt<'a> {
     #[cfg_attr(feature = "serde", serde(borrow))]
+    /// A compound assignment, such as `+=`
+    #[cfg(feature = "roblox")]
+    #[display(fmt = "{}", _0)]
+    CompoundAssignment(CompoundAssignment<'a>),
     /// An assignment, such as `x = 1`
     #[display(fmt = "{}", _0)]
     Assignment(Assignment<'a>),
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -387,10 +387,6 @@ pub enum Value<'a> {
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 pub enum Stmt<'a> {
     #[cfg_attr(feature = "serde", serde(borrow))]
-    /// A compound assignment, such as `+=`
-    #[cfg(feature = "roblox")]
-    #[display(fmt = "{}", _0)]
-    CompoundAssignment(CompoundAssignment<'a>),
     /// An assignment, such as `x = 1`
     #[display(fmt = "{}", _0)]
     Assignment(Assignment<'a>),
@@ -429,6 +425,11 @@ pub enum Stmt<'a> {
     /// Only available when the "roblox" feature flag is enabled.
     #[cfg(feature = "roblox")]
     Continue(Cow<'a, TokenReference<'a>>),
+    /// A compound assignment, such as `+=`
+    /// Only available when the "roblox" feature flag is enabled
+    #[cfg(feature = "roblox")]
+    #[display(fmt = "{}", _0)]
+    CompoundAssignment(CompoundAssignment<'a>),
     /// A type declaration, such as `type Meters = number`
     /// Only available when the "roblox" feature flag is enabled.
     #[cfg(feature = "roblox")]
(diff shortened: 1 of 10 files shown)


GRAMMAR.md
```diff
@@ -61,4 +61,9 @@ call ::= args | ':' Name args
 suffix ::= call | index
 var ::= prefix {suffix} index | Name
 functioncall ::= prefix {suffix} call
+```
+
+Roblox specific grammar
```
Kampfkarren, line 66: We don't really update this file--I think it was mostly for reference. Plus, this isn't correct to begin with since you don't use this anywhere outside of defining it.

full-moon/src/ast/parsers.rs
```diff
@@ -1383,6 +1385,32 @@ cfg_if::cfg_if! {
                 ))
             }
         );
+
+        // Roblox Compound Assignment
+        #[derive(Clone, Debug, Default, PartialEq)]
+        struct ParseCompoundAssignment;
+        define_parser!(
```
Kampfkarren, line 1392: I'm almost certain this needs to be define_roblox_parser. Make sure you test both with and without the Roblox feature flag.

full-moon/src/ast/parsers.rs
```diff
@@ -1383,6 +1385,32 @@ cfg_if::cfg_if! {
...
+
+        // Roblox Compound Assignment
+        #[derive(Clone, Debug, Default, PartialEq)]
+        struct ParseCompoundAssignment;
+        define_parser!(
+            ParseCompoundAssignment,
+            CompoundAssignment<'a>,
+            |_, state: ParserState<'a>| {
+                let (state, lhs) = ParseVar.parse(state)?;
+                let (state, compoundop) = ParseCompoundOp.parse(state)?;
```
Kampfkarren, line 1397: compound_operator

full-moon/src/ast/parsers.rs
```diff
@@ -1383,6 +1385,32 @@ cfg_if::cfg_if! {
...
+                let (state, rhs) = expect!(
+                    state,
+                    ParseExpression.parse(state),
+                    "expected values"
```
Kampfkarren, line 1401: Only one value.

full-moon/src/ast/types.rs
```diff
@@ -302,3 +302,70 @@ impl<'a> TypeSpecifier<'a> {
         &self.type_info
     }
 }
+
+macro_rules! make_op {
```
Kampfkarren, line 306: If this is C&P from where we make BinOp and UnOp, this should be in a utility module somewhere instead.

full-moon/src/ast/types.rs
```diff
@@ -302,3 +302,70 @@ impl<'a> TypeSpecifier<'a> {
...
+        }
+    }
+
+    /// The variable assigned to, the `x` part of `x += 1`
+    pub fn lhs(&self) -> &Var<'a> {
+        &self.lhs
+    }
+
+    /// The operator used, the `+=` part of `x += 1`
+    pub fn compoundop(&self) -> &CompoundOp<'a> {
```
Kampfkarren, line 363: compound_operator

full-moon/src/ast/parsers.rs
JohnnyMorganz: I added this here because it is under the conditional statement (this is also where the type parsers were, which just use define_parser)

full-moon/src/ast/types.rs
JohnnyMorganz: I’m a bit unsure on this. I tried using #[macro_export] above the same code in the original module, then #[macro_use] above the import in this file, and it seemed to error. I’ll dig into this more, and probably move it into a utility file

Kampfkarren: Importing macros is kinda funky, but we define them outside for things like `define_parser!`

JohnnyMorganz: I have moved this into the `parser_utils` module - although I don't know if I should create a separate utility module instead as it isn't *specifically* for the parser

GRAMMAR.md
```diff
@@ -61,4 +61,4 @@ call ::= args | ':' Name args
 suffix ::= call | index
 var ::= prefix {suffix} index | Name
 functioncall ::= prefix {suffix} call
-```
\ No newline at end of file
+```
```
Kampfkarren, line 64: I think you changed a newline here so it's still showing a diff.

full-moon/src/ast/mod.rs
```diff
@@ -387,6 +387,10 @@ pub enum Value<'a> {
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 pub enum Stmt<'a> {
     #[cfg_attr(feature = "serde", serde(borrow))]
+    /// A compound assignment, such as `+=`
+    #[cfg(feature = "roblox")]
+    #[display(fmt = "{}", _0)]
+    CompoundAssignment(CompoundAssignment<'a>),
```
Kampfkarren, line 393: These should be on the bottom, right before where `TypeDeclaration` is.

full-moon/src/ast/parsers.rs
```diff
@@ -349,6 +349,8 @@ define_parser!(
     ParseStmt,
     Stmt<'a>,
     |_, state: ParserState<'a>| parse_first_of!(state, {
+        @#[cfg(feature = "roblox")]
+        ParseCompoundAssignment => Stmt::CompoundAssignment,
```
Kampfkarren, line 353: Same here--put this before `ParseTypeDeclaration` so that all the Roblox parsers are close together.

full-moon/src/ast/parsers.rs
```diff
@@ -1383,6 +1385,32 @@ cfg_if::cfg_if! {
                 ))
             }
         );
+
+        // Roblox Compound Assignment
+        #[derive(Clone, Debug, Default, PartialEq)]
+        struct ParseCompoundAssignment;
```
Kampfkarren, line 1391: Alphabetical order with the rest, so that `ParseCompoundAssignment` is not after `ParseTypeSpecifier`.

Kampfkarren:
Sorry to keep bugging you on this, but Compound comes before Continue. Last change before I merge it. 😛

JohnnyMorganz:
No problem, order should be fixed now

PR #88, opened 2020-09-07
Improve type parsing

### LINKED ISSUE #56: Type declaration of objects do not support trailing comma
```lua
type Foo = {
	foo: any,
}
```

...this is a parse error, when it should not be.

### LINKED ISSUE #86: Should support 'export' keyword for Roblox Lua types
Should support the `export` keyword used in Roblox scripts, documented [here](https://roblox.github.io/luau/typecheck.html#module-interactions).

Example:

```lua
export type Foo = {
   Bar: number
}
```

### LINKED ISSUE #87: When under Roblox feature flag, should support types from other modules
This is valid Luau:
```
local foo = require(ModuleScript)
local bar: foo.type_name = ...
```

It throws a parsing at the `.` in `foo.type_name` though. It shouldn't do that.

JohnnyMorganz:
This PR:
- Adds support for the `export` keyword for types (Closes #86)
This is done by creating a new `ExportedTypeDeclaration` which is based off `TypeDeclaration`
- Allows object type definitions to have trailing commas (Closes #56)
- Adds support for using type definitions from other modules (Closes #87)
This extends `TypeInfo` with a `Module` type. However, I needed to create an `IndexedTypeInfo` to allow support for generics within the module (`local x: module.Array<string>`) - this couldn't just use `TypeInfo` as things like Callbacks and Union/Intersection are not allowed for the index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added support for exporting types (`export type Foo = { bar: number }`) under the `roblox` feature flag
 
+### Fixed
+- Fixed type declaration of objects not supporting trailing commas
+
 ## [0.6.2] - 2020-07-11
 ### Fixed
 - Fixed an error related with `visit_compound_op` and the `roblox` feature flag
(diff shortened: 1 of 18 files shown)


Kampfkarren:
Very nice work!

PR #89, opened 2020-09-07
Move continue to LastStmt from Stmt

### LINKED ISSUE #79: Move continue to LastStmt instead of Stmt
Makes it so you can't do:

```lua
if true then
    continue
    print("more code")
    continue continue continue
end
```

JohnnyMorganz:
Fixes #79

### DIFF
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -98,6 +98,10 @@ pub enum LastStmt<'a> {
     #[cfg_attr(feature = "serde", serde(borrow))]
     /// A `break` statement
     Break(Cow<'a, TokenReference<'a>>),
+    /// A continue statement
+    /// Only available when the "roblox" feature flag is enabled.
+    #[cfg(feature = "roblox")]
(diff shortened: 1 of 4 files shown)


full-moon/src/ast/mod.rs
```diff
@@ -100,6 +100,10 @@ pub enum LastStmt<'a> {
     Break(Cow<'a, TokenReference<'a>>),
     /// A `return` statement
     Return(Return<'a>),
+    /// A continue statement
+    /// Only available when the "roblox" feature flag is enabled.
+    #[cfg(feature = "roblox")]
+    Continue(Cow<'a, TokenReference<'a>>),
```
Kampfkarren, line 106: Alphabetical order, move this between Break and Return.

Kampfkarren (changes requested):

Kampfkarren:
This should appear in the CHANGELOG, you should write it as explaining and fixing #79.

Kampfkarren (approved):

PR #90, opened 2020-09-13
use `str::len` when indexing, `.chars().count()` for `advance` value

### LINKED ISSUE #80: Parsing Errors with complex emojis (UTF-8 with Fitzpatrick modifiers)
To reproduce:
Just enter a comment in bracket format (prefix --[[ ]]) with an emoji that has skin color. Examples:
--[[👨🏾‍💻]]
--[[🤴]]
--[[🧓🏽]]

The line after it will have a parse error. Actually, upon further testing, it throws even more errors, including saying that local variables were not local. 

image

Erutuon:
This at least fixes #80 and the minimized version of the code I was trying to parse
```lua
--[=[μέλλον]=]

--v
```
where `v` was reported as an extra identifier.

The `advance` field in `TokenAdvancement` seems to deal with code points, so it should be created using values created with `.chars().count()`, and indexing strings must be done using byte counts (`.len()`). The two measurements are interchangeable for ASCII strings, so `"--[[]]".chars().count()` is equal to `"--[[]]".len()`, but `"--[[]]".chars().count()` is clearer.


Kampfkarren:
> This at least fixes #80 and the minimized version of the code I was trying to parse

Can you add your code as a test?

Erutuon:
Added my code as well as an emoji example based on issue #80 in two multi-line-comments folders, but do I need to add `ast.json` and `tokens.json` as well?

Erutuon:
Ahh! I see, JSON is added when I run `cargo test`. Added a multi-line string case that failed in the old version.

Kampfkarren:
One last nitpick, can you put this in the CHANGELOG as well? Then it'll be ready for merge.

PR #96, opened 2020-10-10
Add test cases for multi-line quoted string

### LINKED ISSUE #83: Parsing fails when trying to parse a multi-line quoted string
This is valid in Lua 5.1:
```lua
local foo = "bar\
baz"
```

But full-moon panics when it comes across something like this. It shouldn't do that and instead should parse it properly.

zummenix:
~~I want to tackle the #83. So far I've added test cases for that, they are failing right now.~~

~~My next step is to add code to remove hanging `\` at the end of a line in `advance_quote` function of `tokenizer.rs` and save a string in `Token::StringLiteral` as new `Cow::Owned`. What do you think?~~

Closes #83

### DIFF
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -1264,6 +1264,18 @@ mod tests {
         );
 
         test_advancer!(
+            advance_quote("\"hello\\\nworld\""),
+            Ok(Some(TokenAdvancement {
+                advance: 14,
+                token_type: TokenType::StringLiteral {
+                    literal: Cow::from("hello\\\nworld"),
+                    multi_line: None,
+                    quote_type: StringLiteralQuoteType::Double,
+                },
+            }))
+        );
+
+        test_advancer!(
             advance_quote("\"hello"),
             Err(TokenizerErrorType::UnclosedString)
         );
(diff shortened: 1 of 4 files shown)


full-moon/src/tokenizer.rs
```diff
@@ -1263,6 +1263,24 @@ mod tests {
             }))
         );
 
+        test_advancer!(
+            advance_quote(
+                r#""hello\
+world""#
+            ),
```
Kampfkarren, line 1270: Can you indent the `world` to be on the same line?

Kampfkarren:
> My next step is to add code to remove hanging \ at the end of a line in advance_quote function of tokenizer.rs and save a string in Token::StringLiteral as new Cow::Owned. 

I'm not sure what you mean by this, can you elaborate?

zummenix:
@Kampfkarren 

I've made tests to pass. As you can see the problem is that it leaves `\` in the string.

Kampfkarren:
I think it should keep `\` in the string, no? full-moon is intended to be lossless.

It's possible even that the issue report was wrong if you can get these tests to pass without fiddling with the parser at all.

zummenix:
Got it! So, let me clean up the commits. Do you want me to do something else here?

Kampfkarren:
Don't believe so--just change the PR name and such to specify the new purpose of this PR, being to add tests.

Kampfkarren:
Thanks!

PR #104, opened 2020-11-27
Use intra doc links and remove unnecessary links

zummenix:
Use intra doc links feature of the latest `rustc 1.48.0`. I also removed some unnecessary linking to items, for example for methods/functions that return the item which is already linked in the docs. And fixed some warnings reported by `cargo doc` command. 

External `cargo deadlinks` command returns no results now 🥳


Kampfkarren:
Will double check these today, thanks!

Kampfkarren:
Awesome!

PR #112, opened 2020-12-19
Expose Punctuated parameters in FunctionBody

### LINKED ISSUE #109: Expose the parameters in FunctionBody as a Punctuated sequence rather than an iterator
Currently, the only way to access the parameters in FunctionBody is through `iter_parameters()`, which returns an Iterator over the `Parameters`, ignoring any punctuation in the Punctuated sequence
This means there is no way to access the punctuation, which causes an isssue here:
```lua
function foo(
    bar, -- foo
    baz --[[ bar ]]
)
    print("test")
end
```
where the comments cannot be retrieved.

FunctionBody should have a function `parameters() -> Punctuated<'a, Parameter<'a>>` so that the punctuation can be accessed.

`iter_parameters()` can also be deprecated in favour of `Punctuated.iter()`

JohnnyMorganz:
This PR adds a `.parameters()` method to `FunctionBody` in order to retrieve the Punctuated sequence of parameters
Closes #109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Added `with_XXX` methods to Roblox-related structs under the `roblox` feature flag
+- Added support for retrieving the `Punctuated` sequence of parameters in a `FunctionBody`
 
 ### Changed
 - Use intra doc links, remove unnecessary linking for some items in docs.
-- Added support for retrieving the `Punctuated` sequence of parameters in a `FunctionBody`
 - `FunctionBody::iter_parameters` is now deprecated in favour of `punctuated().iter` 
 
 ## [0.7.0] - 2020-11-06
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added support for retrieving the `Punctuated` sequence of parameters in a `FunctionBody`
```
Kampfkarren, line 8: Put this under ### Added.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- `FunctionBody::iter_parameters` is now deprecated in favour of `Punctated::iter`
```
Kampfkarren, line 9: Put this under ### Changed.

Kampfkarren: Also, your comment isn't accurate--it's deprecated in favor of `parameters().iter`, not `Puncutated::iter` generally.

Kampfkarren (changes requested):

CHANGELOG.md
Kampfkarren: Not resolved--this needs to be under ### Added, and not ### Changed 😛

JohnnyMorganz: Whoops!

Kampfkarren:
:+1: Will merge once CI passes, thanks!

PR #114, opened 2020-12-19
Add support for types in for loops

### LINKED ISSUE #103: Types in for parameters don't work.
image
This is not supported by full-moon but in Roblox Studio works perfectly:
image
Both parameters can have a type.

It also should work in numeric fors although is redundant.
image

JohnnyMorganz:
Adds type_specifiers for Generic and Numeric for loops under the `roblox` feature flag
Fixes #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Added support for types within generic and numeric for loops under the `roblox` feature flag
 
 ### Changed
 - Use intra doc links, remove unnecessary linking for some items in docs.
--- a/full-moon/src/ast/parsers.rs
+++ b/full-moon/src/ast/parsers.rs
@@ -459,9 +459,26 @@ define_parser!(
     ParseNumericFor,
     NumericFor<'a>,
     |_, state: ParserState<'a>| {
-        let (state, for_token) = ParseSymbol(Symbol::For).parse(state)?;
-        let (state, index_variable) =
-            expect!(state, ParseIdentifier.parse(state), "expected names");
+        let (mut state, for_token) = ParseSymbol(Symbol::For).parse(state)?;
+
+        let index_variable;
+        let mut type_specifier = None;
+
+        if cfg!(feature = "roblox") {
+            let (new_state, (new_index_variable, new_type_specifier)) =
+                expect!(state, ParseNameWithType.parse(state), "expected names");
+
+            state = new_state;
+            index_variable = new_index_variable;
+            type_specifier = new_type_specifier;
+        } else {
+            let (new_state, new_index_variable) =
+                expect!(state, ParseIdentifier.parse(state), "expected names");
+
+            state = new_state;
+            index_variable = new_index_variable;
+        }
+
         let (state, equal_token) = ParseSymbol(Symbol::Equal).parse(state)?; // Numeric fors run before generic fors, so we can't guarantee this
         let (state, start) = expect!(
             state,
@@ -512,6 +529,8 @@ define_parser!(
                 do_token,
                 block,
                 end_token,
+                #[cfg(feature = "roblox")]
+                type_specifier,
             },
         ))
     }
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Added support for types within generic and numeric for loops under the `roblox` feature flag
```
Kampfkarren, line 8: Put this under ### Added.

full-moon/src/ast/parsers.rs
```diff
@@ -523,12 +542,37 @@ define_parser!(
...
-            state,
-            OneOrMore(ParseIdentifier, ParseSymbol(Symbol::Comma), false).parse(state),
-            "expected names"
-        );
+        let (mut state, for_token) = ParseSymbol(Symbol::For).parse(state)?;
+
+        let mut names;
+        let mut type_specifiers = Vec::new();
+
+        if cfg!(feature = "roblox") {
```
Kampfkarren, line 550: I'm pretty sure you need `cfg_if` for this instead of `if cfg!`--otherwise, you're going to be referencing fields that don't exist, leading to a compile error. Make sure to compile this with `--no-default-features` and see if it still works.

full-moon/src/ast/parsers.rs
```diff
@@ -523,12 +542,37 @@ define_parser!(
...
     |_, state: ParserState<'a>| {
-        let (state, for_token) = ParseSymbol(Symbol::For).parse(state)?;
-        let (state, names) = expect!(
-            state,
-            OneOrMore(ParseIdentifier, ParseSymbol(Symbol::Comma), false).parse(state),
-            "expected names"
-        );
+        let (mut state, for_token) = ParseSymbol(Symbol::For).parse(state)?;
+
+        let mut names;
```
Kampfkarren, line 547:
```suggestion
        let names;
```

Variables that are assigned later, and only once, do not need to be marked mutable.

full-moon/src/ast/visitors.rs
```diff
@@ -431,3 +431,137 @@ impl<'a> VisitMut<'a> for LocalAssignment<'a> {
         self
     }
 }
+
+impl<'a> Visit<'a> for GenericFor<'a> {
+    fn visit<V: Visitor<'a>>(&self, visitor: &mut V) {
+        visitor.visit_generic_for(self);
+        self.for_token.visit(visitor);
+
+        let mut type_specifiers;
```
Kampfkarren, line 440:
```suggestion
        let type_specifiers;
```

full-moon/src/ast/visitors.rs
```diff
@@ -431,3 +431,137 @@ impl<'a> VisitMut<'a> for LocalAssignment<'a> {
...
+        visitor.visit_generic_for_end(self);
+    }
+}
+
+impl<'a> VisitMut<'a> for GenericFor<'a> {
+    fn visit_mut<V: VisitorMut<'a>>(mut self, visitor: &mut V) -> Self {
+        self = visitor.visit_generic_for(self);
+        self.for_token = self.for_token.visit_mut(visitor);
+
+        let mut type_specifiers;
```
Kampfkarren, line 473:
```suggestion
        let type_specifiers;
```

Kampfkarren (changes requested):

full-moon/src/ast/visitors.rs
JohnnyMorganz: This needs to be labelled as `mut` I think because `next()` is being called on the iterator later, mutating it

full-moon/src/ast/visitors.rs
JohnnyMorganz: Same as above, the iterator is being mutated

full-moon/src/ast/parsers.rs
JohnnyMorganz: This needs to be labelled `mut` as well because the Punctuated struct is being mutated later when the names are pushed to it

full-moon/src/ast/parsers.rs
JohnnyMorganz: Hm, I just copied what was used in ParseLocalAssignment 😛, but I can look into it

Kampfkarren: In that case, don't bother--I forgot out CI does --no-default-features, which seems to pass anyway.

PR #118, opened 2020-12-21
Expose Punctuated fields in TableConstructor

JohnnyMorganz:
This PR adds a `.fields()` method to FunctionBody in order to retrieve the Punctuated sequence of fields
Deprecates `FunctionBody::iter_fields`


Kampfkarren:
Thanks!

PR #124, opened 2021-02-19
Add support for accessing semicolon for statements

### LINKED ISSUE #123: No access to semicolon at the end of statements
Currently, you are able to add in a semicolon token using the `with_stmts` function in the Block node, however the iterator over statements does not provide access to the semicolon token.

This means it is impossible to access this token, and therefore its information is lost.

JohnnyMorganz:
Closes #123


Kampfkarren:
Thanks!

PR #127, opened 2021-03-02
Update to nom 6.

tomprince:


full-moon-derive/src/symbols.rs
```diff
@@ -105,33 +122,16 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
             }
         }
 
-        fn parse_symbol(code: &str) -> IResult<&str, &str> {
-            let combinator = alt((
+        fn parse_symbol(code: &str) -> IResult<&str, Symbol> {
+            use nom::character::complete::alphanumeric1;
+            use nom::combinator::{not, value};
+            use nom::sequence::{terminated};
```
Kampfkarren, line 128: Move these out to the beginning (before `use indexmap::IndexMap`), and structure it like `use nom::{character::complete::alphanumeric1, combinator::{not, value}, sequence::terminated}`. Rustfmt should help with this.

full-moon-derive/src/symbols.rs
```diff
@@ -62,17 +62,34 @@ impl Parse for SymbolsInput {
...
 
-    let string = symbols.values().collect::<Vec<_>>();
-    let splits = string.chunks(ALT_LIMIT).map(|string| {
+    let string: Vec<_> = symbols.values().collect();
+    let ident: Vec<_> = symbols.keys().collect();
+    let symbols: Vec<_> = symbols.iter().collect();
+    let splits = symbols.chunks(ALT_LIMIT).map(|symbols| {
+        let parsers = symbols.iter().map(|(symbol, string)| {
+            let matcher = if string.value().chars().all(
+                //TODO: ensure first character is not alphanumeric
```
Kampfkarren, line 71: Can't you just get the first char with `let first_char = string.value().chars().next()`?

What happens if the first character is not alphanumeric?

Kampfkarren (changes requested):
Thanks!

full-moon-derive/src/symbols.rs
tomprince: I've adjusted these to have the nested use structure, but putting them at the top of the file won't work, as this is embedded in a macro.

full-moon-derive/src/symbols.rs
tomprince: The idea is we have two type of symbols, those that are identifier-like, which are terminated by a non-identifier character. I could definitely check for the first character, to see if it is a digit. I didn't bother, partly because there are no symbols that begin with a digit anyway (and thinking about it more, I'm not even sure what the right behavior is).

I wonder if it makes sense to just change this to `is_ascii_alphabetic` and leave a note about handling other cases if it becomes required. (I've ended up doing this + `_`.)

full-moon-derive/src/symbols.rs
Kampfkarren: Good point, my mistake.

Kampfkarren:
Thanks!

PR #131, opened 2021-03-02
Make tokens at the begining of a new line start on that line.

### LINKED ISSUE #128: Start position of tokens at the beginning of the line are recorded as being at the end of the previous line.
I was experimenting with using [nom_locate](https://github.com/fflorent/nom_locate/) for tracking source positions. I've got a mostly working branch with that, but I discovered that it generates different source positions in a number of places:
- any token at the beginning of the line currently has its position as being at the end of the previous line
- a whitespace token consisting of a single new line has start and end position with the same line and column, on the previous line

I think either decision for the second difference could be argued to be correct, but I think the first one is definitely an error. I've got a patch to make the first change (and am working on updating the tests), but I'm wondering if the second change also wants to be made.

tomprince:
Fixes #128.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated dependency cfg_if to v1.0
 
 ### Fixed
-- Fixed the start position of tokens at the beginnig of a line to not be at the end of the previous line.
+- Fixed the start position of tokens at the beginning of a line to not be at the end of the previous line.
 
 ## [0.9.0] - 2020-12-21
 ### Added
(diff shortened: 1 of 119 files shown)


Kampfkarren:
Can you update the CHANGELOG as well, under an "Unreleased" category if one does not exist already?

CC @JohnnyMorganz, if you are interested or have any comments with regards to this.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Updated dependency cfg_if to v1.0
 
+### Fixed
+- Fixed the start position of tokens at the beginnig of a line to not be at the end of the previous line.
```
JohnnyMorganz, line 15:
```suggestion
- Fixed the start position of tokens at the beginning of a line to not be at the end of the previous line.
```

JohnnyMorganz (commented):
Looks good to me, I have nothing to add

Kampfkarren:
Thanks!

Issue #129, opened 2021-03-02
Kill the `Cow`s and lifetimes


Kampfkarren:
From what I've heard from other developers trying to use full-moon (Dreamcraft), the lifetime juggling everywhere is a big mess. I can attest to the same with Selene. It leads to a lot of weird code like matching `&*node` and such, and is just generally confusing.

The motivation behind `<'ast>` was that full-moon should be able to borrow from the string provided. Furthermore, the `Cow`s are necessary in order to allow both borrowed (from the string) and owned (created from mutation).

However, this is not the way it works right now. I could never get borrowed to work, so it just ends up cloning everything anyway, and in really awful news, it led to an `unsafe` being created in order to get the dang thing to work. While it's likely both these issues are solvable, this compounded with the hard to use API means I would rather just kill 'ast and the Cows.

Instead, full-moon will simply just own everything it has, and not try to bother borrowing.

tomprince:
Looking at the code, the only place where `Ast.tokens` field is used is for access to the EOF token, so there is no reason post-parse, to not keep the tokens (or [parts of them](https://github.com/Kampfkarren/full-moon/issues/23#issuecomment-799652220)) directly in the ast.

---

The one reason we might not want to do that during parsing, we copy `TokenReference`s a bunch when backtracking. One idea I had (though it conflicts with my suggestions in #23) was to parameterize all the ast nodes with the type of token, which could then be `Token<'text>` or `TokenReference<'text>` or `Cow<'parse, TokenReference<'text>>` and we could use the last during parsing, and return one of the former two depending on the use-case.

Given my suggestion in #23, another thought I had is storing trivia as a pair of (wrapped) `Cow<'text, str>`s, and reparsing as required. This would make trivia cheap to copy during parsing, at the cost of making manipulating it more expensive to manipulate them later. Or we could maybe use a [smallvec](https://docs.rs/smallvec) of leading and trailing trivia.

----

All that being said, if performance is adequate right now, with copying everything always with `Vec`s of trivia, maybe it isn't worth worrying about this kind of optimization right now.

Kampfkarren:
Like I said before, I can't remember the motivation behind the tokens field in general other than that I wanted to make `print` implemented as just iterating over every token in the vector and printing it. That ship has sailed.

tomprince:
[...]

PR #132, opened 2021-03-03
Remove unsafe.

tomprince:
I saw  #129, and was curious if it was possible to remove the use of unsafe. This manages to do that. I've not yet figured out how to remove the clone in peek, but I'm investigating.

### DIFF
--- a/full-moon/src/ast/parser_util.rs
+++ b/full-moon/src/ast/parser_util.rs
@@ -101,10 +101,17 @@ macro_rules! make_op {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! define_parser {
+    ($parser:ident, $node:ty, |_, $state:ident| $body:expr) => {
+        define_parser! {$parser, $node, |_, mut $state: ParserState<'a>| $body}
+    };
+    ($parser:ident, $node:ty, |$self:ident, $state:ident| $body:expr) => {
+        define_parser! {$parser, $node, |$self:&$parser, mut $state: ParserState<'a>| $body}
+    };
     ($parser:ident, $node:ty, $body:expr) => {
         impl<'a> Parser<'a> for $parser {
             type Item = $node;
 
+            #[allow(unused_mut)]
             fn parse(
                 &self,
                 state: ParserState<'a>,
@@ -400,4 +407,4 @@ where
 #[derive(Clone, Debug, PartialEq)]
 pub struct NoDelimiter;
 
-define_parser!(NoDelimiter, (), |_, state: ParserState<'a>| Ok((state, ())));
+define_parser!(NoDelimiter, (), |_, state| Ok((state, ())));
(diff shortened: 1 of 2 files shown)


full-moon/src/ast/parser_util.rs
```diff
@@ -39,27 +39,21 @@ impl<'a> ParserState<'a> {
         }
     }
 
-    // TODO: This is super bad, containing both unsafe code and a mandatory clone
-    // on every call so that everything is backwards compatible, since it SHOULD
-    // just borrow. It is only like this because of a failure to tackle lifetimes.
+    // TODO: This is bad, containing both a mandatory clone on every call so that everything is
```
Kampfkarren, line 42:
```suggestion
    // TODO: This is bad, containing a mandatory clone on every call so that everything is
```

🎉

full-moon/src/ast/parser_util.rs
```diff
@@ -39,27 +39,21 @@ impl<'a> ParserState<'a> {
...
         }
 
-        let result = unsafe {
-            &*self
-                .tokens
-                .add(self.index)
-                .as_ref()
-                .expect("couldn't peek, no eof?")
-        };
+        let result = { self.tokens.get(self.index).expect("couldn't peek, no eof?") };
```
Kampfkarren, line 50:
```suggestion
        let result = self.tokens.get(self.index).expect("couldn't peek, no eof?");
```

Kampfkarren:
Awesome!!! 😄

tomprince:
Looking more into this, it appears that the underling issue with needing the clone is tied to the fact that we want references from `ast.nodes` to `ast.tokens`, which rust does not allow (although there are crates like [ouroboros](https://docs.rs/ouroboros/) which can provide a safe wrapper around a struct which does that).

So I'm not sure how much value there is in making this change as-is, if #129 is going to be implemented. I've not yet used this package [1], so I'm not sure what the troubles there are from. I would be happy to investigate what would be required to implement something using `ouroboros`, if you do end up wanting to keep the zero-copy implementation. I think the only thing that couldn't be preserved is `Ast::with_eof`. I don't know the history of using `Cow`, but I suspect that it would be possible to replace them with plain references, though I'm not sure how that would interact with `Ast::nodes_mut`,

[1] I'm interested in re-implementing https://github.com/Benjamin-Dobell/luabundle in rust, but need to add some lua 5.2/moonlua support before that is valuable to me.

tomprince:
Thinking about this further, and looking over the code, I'm wondering if there is any reason to keep the tokens store separately from the ast? I've not looked exhaustively, but it doesn't not appear like tokens are shared between nodes. In which case; at least after parsing is complete, there is no reason for not storing token's inline in the ast[1], though handling backtracking gracefully might keeping references there.

[1] This is separate from having the text that `Token` points to not be shared.

Kampfkarren:
Honestly, a lot of the early API decisions like this weren't made with any great reason, other than that I wanted something that works!

Why are tokens separate from AST? I don't remember, I think the plan was to have `print` be implemented just by going through that vector of tokens and joining them, but that ship has sailed, and `print` already exists by just recursively visiting anything.

tomprince:
On reflection, it probably makes sense to land this now, rather than waiting for a better solution, particularly as I managed to make `define_parser` smart enough to not need to explicitly write type of `ParserState` in invocations.

Kampfkarren:
Thanks!

PR #134, opened 2021-03-04
TokenReference PartialEq now checks if leading_trivia and trailing_trivia are equal

Kampfkarren:
This was causing tests to not fail when trivia was different.

Will be merged once it is confirmed selene did not rely on this in some way.


Kampfkarren:
Selene tests pass.

PR #135, opened 2021-03-04
Implement new binary operator AST

### LINKED ISSUE #43: Rename ParseExpression to ParenthesesExpression or something similar
Weird internal(?) name.

### LINKED ISSUE #64: Needs precedence rules
Precedence rules are missing for operators.
Example:
```Lua
local a = 1 == 1 and 0
```
Is parsed as
```Lua
local a = 1 == (1 and 0)
```
and expected behavior is
```Lua
local a = (1 == 1) and 0
```

JohnnyMorganz:
This PR modifies the way binary operators are parsed and handled in the AST.
Previously, binary operators were handled using `BinOpRhs` and a `binop` member in `Expression::Value{}`. This did not have support for precedence.
This PR is a **breaking change**

Changes made:
- Added a new struct enum `Expression::BinaryOperator { lhs, binop, rhs }`. The `binop` member from `Expression::Value` has been removed.
- Removed the `BinOpRhs` struct
- Added a new method `operator.precedence()` available on `BinOp` and `UnOp` - this returns the precedence value for that operator, on a scale of 1 to 8, where 8 is the highest precedence
- Added a new method `binop.is_right_associative()` to determine if a BinOp is right associative
- ~~Removed `Value::ParseExpression` as it is no longer required. This was used when we had a LHS of a binop that was contained within parentheses. This is no longer required, as it is now part of `lhs` in `Expression::BinaryOperator` - the `lhs` will return an `Expression::Parentheses`~~ (See: https://github.com/Kampfkarren/full-moon/pull/135#issuecomment-790821559)
- Removed the `visit_bin_op`/`visit_bin_op_end` and the related VisitorMut visitors. This is now handled within the expression visitor.

Closes #64 

Old tests have been verified to still be correct. I also just want to verify that you are happy with the new API names

### DIFF
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -262,49 +262,23 @@ impl Default for TableConstructor<'_> {
     }
 }
 
-/// A binary operation, such as (`+ 3`)
-#[derive(Clone, Debug, Display, PartialEq, Owned, Node, Visit)]
-#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
-#[display(fmt = "{}{}", bin_op, rhs)]
-#[visit(visit_as = "bin_op")]
-pub struct BinOpRhs<'a> {
-    #[cfg_attr(feature = "serde", serde(borrow))]
-    bin_op: BinOp<'a>,
-    rhs: Box<Expression<'a>>,
-}
-
-impl<'a> BinOpRhs<'a> {
-    /// Creates a new BinOpRhs from the given binary operator and right hand side
-    pub fn new(bin_op: BinOp<'a>, rhs: Box<Expression<'a>>) -> Self {
-        Self { bin_op, rhs }
-    }
-
-    /// The binary operation used, the `+` part of `+ 3`
-    pub fn bin_op(&self) -> &BinOp<'a> {
-        &self.bin_op
-    }
-
-    /// The right hand side of the binary operation, the `3` part of `+ 3`
-    pub fn rhs(&self) -> &Expression<'a> {
-        self.rhs.as_ref()
-    }
-
-    /// Returns a new BinOpRhs with the given binary operator token
-    pub fn with_bin_op(self, bin_op: BinOp<'a>) -> Self {
-        Self { bin_op, ..self }
-    }
-
-    /// Returns a new BinOpRhs with the given right hand side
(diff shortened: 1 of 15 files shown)


JohnnyMorganz:
Turns out `Value::ParseExpression` was more important than I realised. Without it, something like `(origin - character.PrimaryPart.Position).Unit` will break the parser. Currently, if we parse ParensExpression inside of ParseExpression, it will take the expression within parentheses and throw a leftover token error when it reaches `.Unit`, rather than ignoring it in ParseExpression and leaving it for ParseVarExpression/ParsePrefix

I'm wondering, is there any way we can fix this? I'm assuming you have tried before. The currently layout seems a bit verbose (whenever we have `Expression::Parentheses`, we actually have `Expression::Value{ value = Value::ParseExpression(Expression::Parentheses) }`)

full-moon/src/ast/mod.rs
```diff
@@ -2060,6 +2024,33 @@ make_op!(BinOp,
...
+            | BinOp::GreaterThanEqual(_)
+            | BinOp::LessThanEqual(_)
+            | BinOp::TildeEqual(_)
+            | BinOp::TwoEqual(_) => 3,
+            BinOp::And(_) => 2,
+            BinOp::Or(_) => 1,
+        }
+    }
+
+    /// Whether the operator is right associative. If not, it is left associative
```
Kampfkarren, line 2047:
```suggestion
    /// Whether the operator is right associative. If not, it is left associative.
```

Kampfkarren (approved):
The new types are fine, yeah.

Kampfkarren:
The reason ParseExpression is the way it is now is simply because I was copying the GRAMMAR guide from Lua, which has that level of completeness. I haven't tried to fix that, though it feels out of scope for this PR, no?

JohnnyMorganz:
Originally I though ParseExpression was a side effect of how BinOps were originally setup - `(foo) and bar`, and lhs = value, not expression.

But seeing as it's actually used for other things too, then yeah it's probably out of scope here

Kampfkarren:
> happy to move it into a separate PR if you want

Yep, let me know when this is ready.

JohnnyMorganz:
@Kampfkarren Should be ready now (moved the rename to #137)

PR #138, opened 2021-03-04
Luau: Change type assertion operator

### LINKED ISSUE #136: Luau: add support for the new assertion operator
`x :: number` instead of `x as number` - also need to check to see if `as` is still usable for assertion.

JohnnyMorganz:
This PR changes the type assertion operator from `as` to `::`
The `AsAssertion` struct is renamed to `TypeAssertion`
`as_assertion.as_token()` is changed to `type_assertion.assertion_op()`
`::` is now added as a symbol, and `as` is no longer allowed

Closes #136


full-moon/src/ast/types.rs
```diff
@@ -273,31 +273,34 @@ pub enum TypeFieldKey<'a> {
 /// A type assertion using `as`, such as `as number`.
```
Kampfkarren, line 273: Change this comment.

PR #142, opened 2021-03-05
Implements goto and label syntax in Lua 5.2

### LINKED ISSUE #92: Add support for labels/goto
As per the discussion in issue https://github.com/Kampfkarren/selene/issues/154, support needs to be added for `goto`/`::label::` syntax, probably behind a `luajit` crate flag.

This is primarily for LuaJIT, but this is one of the major changes in Lua 5.2 (the other 'breaking' changes are empty statements and `break` being allowed anywhere in the block), so that's worth taking into consideration -- it would probably be trivial after this to support 5.2 entirely.

This issue is mostly meant for the sake of posterity, as the proposal was already accepted on the Selene repo, I just want a record of it here too.

JohnnyMorganz:
This PR begins the support of Lua 5.2, with the introduction of goto and label syntax.
Lua 5.2 support is available behind the `lua52` feature flag [#92 mentioned `luajit` feature flag, but since this is a Lua 5.2 change, I used `lua52` instead - LuaJIT adds goto support in, but they add it as a LastStmt, rather than a Stmt]

This begins by adding two new structs: `Goto` and `Label`. These are both statements, and therefore are added to the `Stmt` enum.
A `define_lua52_parser` macro was also added for parsing Lua 5.2 related syntax.

`pass_cases` and `fail_cases` were both extended to support snapshot testing of Lua 5.2 syntax. Some pass and parser fail test cases have been added to `tests/lua52_cases/`

Closes #92


Kampfkarren:
Can you add this to the `package.metadata.docs.rs` entry as well?

Kampfkarren:
Also, can you add a new entry in test.yml to test Lua 5.2 features?

PR #144, opened 2021-03-06
Fixes UnclosedString when providing an empty string

Kampfkarren:
CC @tomprince -- I'm not super familiar with nom, does this solution look alright?


tomprince (approved):
Yeah. It would probably make sense to leave a link to https://github.com/Geal/nom/issues/1118 in a comment.

PR #146, opened 2021-03-08
New, more consistent names (remove iter_*, and *_list)

### LINKED ISSUE #35: Make API names consistent
For example, `GenericFor::names` and `LocalAssignment:name_list`

Kampfkarren:
The functions were not marked deprecated, as we are ideally moving to 1.0.0 very shortly (and the deprecated stuff should get removed around then anyway).

Closes #35


JohnnyMorganz:
Just a few more that I would point out:
- `LocalFunction.func_body()` vs `FunctionDeclaration.body()`
- `FunctionBody.parameters_parentheses()` - a bit verbose? But I don't mind either way

Also maybe a few things in Luau need a quick glance over?
`type_info` vs `type_definition`, and I think TypeAssertion uses `cast_to`

Kampfkarren:
Good calls, I'll look when I can.

Kampfkarren:
I think `parameters_parentheses` is fine, I'm not really sure how to make it more verbose.

Kampfkarren:
After taking a look at the Luau stuff, I think it's fine as well.

PR #148, opened 2021-03-15
Remove non_exhaustive feature, make everything non_exhaustive instead

Kampfkarren:
I've been told that what we were doing is a violation of features--features are meant to be able to be combined together without breaking compilation.

I'm still unhappy about this solution, as Lua 5.1 users won't ever experience changes, but it makes sense to apply it everywhere so that multiple crates can import full-moon without clashing.

Perhaps we can find a way to split off stuff like Roblox parsing into its own crate that peers with full_moon, Idunno.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated dependency cfg_if to v1.0
 - **[BREAKING CHANGE]** Moved binary operations to `Expression::BinaryOperator`. `binop` has been removed from `Expression::Value`
 - **[BREAKING CHANGE]** Renamed `Value::ParseExpression` to `Value::ParenthesesExpression`
-- **[BREAKING CHANGE (for `roblox` users)]** When using the `non-exhaustive` feature flag (on by default when using the `roblox` feature flag), enums will now be marked as `non_exhaustive`, meaning matches on them must factor in the `_` case.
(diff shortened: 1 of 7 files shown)


JohnnyMorganz:
I guess the decision that needs to be made is whether you think that Luau syntax is still going to change often - it seemed to be quite stable, but I guess the recent change to as assertion showed otherwise.

Like I mentioned earlier, I personally don't like the non exhaustive catch all (I like handling everything explicitly, and I feel as though I might miss something if I'm not careful enough when updating - it might just get caught by the wildcard pattern), but again, I understand the reasoning.
The other option would be to keep bumping the major version when necessary, but is this any better? (depends on how often you think it will happen)

The final option could also be to leave 1.0 to a later date, when we know Luau changes are unlikely, but I'm not sure when that will happen.

Ultimately, go with whatever you think is best, and I'm happy to deal with either way

Kampfkarren:
This is sort of necessary even if I was comfortable constantly bumping the version number and if Luau was stable--if you have two crates with full_moon, and one had roblox/lua52 on, then without non_exhaustive the one without those features could fail to compile, which is against the idea of features.

JohnnyMorganz:
> This is sort of necessary even if I was comfortable constantly bumping the version number and if Luau was stable--if you have two crates with full_moon, and one had roblox/lua52 on, then without non_exhaustive the one without those features could fail to compile, which is against the idea of features. 

Ah I did not think about this! Alright, I guess I'm OK with this then 👍

PR #149, opened 2021-03-15
Only store the EOF token in the ast.

tomprince:
Since we don't refer to any of the other items in the `Vec` after consrtucting
an ast, there is no reason to keep around the rest of the elements.

This does mean that all of the `Cow<'a, TokenReference<'a>` don't have anything
to borrow from, but as we already cloning everything, this isn't a change.

We should either make the borrows work, or remove the use of `Cow` in the API.
See #129 for some discussion of options. In the mean time, this saves some
memory without impacting the API.

### DIFF
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -2119,7 +2119,7 @@ impl<'a> std::error::Error for AstError<'a> {}
 #[derive(Clone, Debug, Owned)]
 pub struct Ast<'a> {
     pub(crate) nodes: Block<'a>,
-    pub(crate) tokens: Vec<TokenReference<'a>>,
+    pub(crate) eof: TokenReference<'a>,
 }
 
 impl<'a> Ast<'a> {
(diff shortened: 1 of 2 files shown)


full-moon/src/ast/mod.rs
```diff
@@ -2152,7 +2152,7 @@ impl<'a> Ast<'a> {
                         stmts: Vec::new(),
                         last_stmt: None,
                     },
-                    tokens,
+                    eof: tokens.pop().unwrap(),
```
Kampfkarren, line 2155: Can you change the `unwrap` to `expect` with a clear error message that there was no EOF?

full-moon/src/ast/mod.rs
```diff
@@ -2166,7 +2166,7 @@ impl<'a> Ast<'a> {
                     if state.index == tokens.len() - 1 {
                         Ok(Ast {
                             nodes: block,
-                            tokens,
+                            eof: tokens.pop().unwrap(),
```
Kampfkarren, line 2169: Here, too.

full-moon/src/ast/mod.rs
tomprince: Following the example between these two instances, I've added a comment on why it should never panic. I could still change it to an `.expect` if you prefer.

Kampfkarren: I don't think it'll ever panic in practice, but it's good to have the `expect` anyway.

Kampfkarren:
Thanks!

Issue #145, opened 2021-03-07
Switch testing to insta with YML snapshots


Kampfkarren:
insta does what we're doing but with a simpler API, more features (like reviewing), and support for YML, which diffs a lot better.

JohnnyMorganz:
I believe this issue was resolved in #150

PR #150, opened 2021-03-16
Prototype of how to generate insta snapshots from the exisiting saved data.

tomprince:
See #145.

The second commit in this PR adds snapshots that use the expected value of the current tests. The third commit then switches to using insta against the test value. 

1. Run the tests with the second commit, commit all the created snapshot files.
2. Run the tests with the third commit, verifying that no snapshot needs review.
3. Remove all snapshot files and rerun tests, to ensure that expression metadata is updated with real test expressions.[1] Commit new snapshots after verifying only metadata changed.

[1] insta apparently store metadata about the expression passed to the assertion macros, but (reasonably) doesn't consider that to require updating the snapshot. I wanted to get the real expression in the files that end up in the repo, which required recreating the snapshots.

----

I went through some work to get the same directory layout, rather than the default one that insta provides. This made it easier to compare with the existing data (one test I did was to use `assert_json_snapshot!` and compare the snapshots with the json files, to verify that the difference was adding metadata), and didn't require moving the test case sources.

If we want to go with more default settings, we'd need to rename `tests/pass/<name>/source.lua` to `tests/pass/<name>.lua`, as insta just use the base name of the file in the snapshot name by default. Even if we do that, I think there is some value in keeping the snapshots next to the source files, so we probably want to customize the name anyway.


tomprince:
It also occurs to me that we could use `settings.set_snapshot_suffix` to handle testing base tests against features, without running into the issue that https://github.com/Kampfkarren/full-moon/blob/6a31907db160dd1c7ea463ee35c46ec7b0035d49/full-moon/tests/pass_cases.rs#L94 was designed to avoid.

tomprince:
I think this is ready for review now. I also created https://github.com/tomprince/full-moon/commit/insta-json-diff which use json rather than yaml, and included a command there to verify that the generated data is identical to the current data.

> It also occurs to me that we could use `settings.set_snapshot_suffix` to handle testing base tests against features, without running into the issue that 
> https://github.com/Kampfkarren/full-moon/blob/6a31907db160dd1c7ea463ee35c46ec7b0035d49/full-moon/tests/pass_cases.rs#L94
> was designed to avoid.

I had a version of this that implemented suffixes based on the set of features, but then I realized that the only source of differences with roblox/luau's type specifiers. I think it would make sense to restructure how those are represented, so that they are not stored as parallel vectors to the names they are associated with (which could then be out of sync).

full-moon/tests/common/mod.rs
```diff
@@ -0,0 +1,18 @@
+use std::fs;
+use std::path::Path;
+
+fn with_insta_settings(path: &Path, f: impl FnOnce()) {
+    let mut settings = insta::Settings::clone_current();
+    settings.set_prepend_module_to_snapshot(false);
+    settings.set_snapshot_path(path);
+    settings.set_input_file(path);
+    settings.bind(|| f())
```
Kampfkarren, line 9: Does this not work with `settings.bind(f())`?

full-moon/src/tokenizer.rs
```diff
@@ -587,8 +587,8 @@ impl<'ast> VisitMut<'ast> for TokenReference<'ast> {
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
 pub struct Position {
     pub(crate) bytes: usize,
-    pub(crate) character: usize,
     pub(crate) line: usize,
+    pub(crate) character: usize,
```
tomprince, line 591: I added this change so that serialization is easier to read (having column then line number is confusing). And in this PR, so we don't have churn of all the `.snap` files.

full-moon/tests/common/mod.rs
tomprince: It does. A previous iteration had been passing `path` to it, which needed the lambda. Fixed.

PR #151, opened 2021-03-18
Get rid of `Cow<'_, TokenReference<'_>>` use.

### LINKED ISSUE #129: Kill the `Cow`s and lifetimes

tomprince:
Fixes #129.

I've not tested this against a downstream consumer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ See [#125](https://github.com/Kampfkarren/full-moon/pull/125) for more details.
 	- `VarExpression::iter_suffixes` -> `VarExpression::suffixes`
 	- `FunctionCall::iter_suffixes` -> `FunctionCall::suffixes`
 	- `FunctionBody::func_body` -> `FunctionBody::body`
-- **[BREAKING HANGE]** All uses of `Cow<'a, TokenRefernce<'a>>` have been removed, though `Cow<'a, str>` remain. 
+- **[BREAKING CHANGE]** All uses of `Cow<'a, TokenReference<'a>>` have been removed, though `Cow<'a, str>` remain. 
 
 ### Fixed
 - Fixed the start position of tokens at the beginning of a line to not be at the end of the previous line.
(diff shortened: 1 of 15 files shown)


Kampfkarren:
This is good work. It still leaves the lifetime, but I think this was the biggest problem with its use.

I'll ask the people who worked on the consumer that requested this in specific.

In the meantime, can you put this in CHANGELOG.md?

tomprince:
> This is good work. It still leaves the lifetime, but I think this was the biggest problem with its use.

Yeah, I think the `Cow<'text, str>` pointing at the original source string makes sense. As https://github.com/Kampfkarren/selene/pull/211 and https://github.com/JohnnyMorganz/StyLua/pull/83 demonstrate, you often don't need to clone the source text (even currently). These `Cow`s were never actually a `Cow::Borrowed` though, so were pure noise.

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ See [#125](https://github.com/Kampfkarren/full-moon/pull/125) for more details.
 	- `VarExpression::iter_suffixes` -> `VarExpression::suffixes`
 	- `FunctionCall::iter_suffixes` -> `FunctionCall::suffixes`
 	- `FunctionBody::func_body` -> `FunctionBody::body`
+- **[BREAKING HANGE]** All uses of `Cow<'a, TokenRefernce<'a>>` have been removed, though `Cow<'a, str>` remain.
```
Kampfkarren, line 41:
```suggestion
- **[BREAKING CHANGE]** All uses of `Cow<'a, TokenRefernce<'a>>` have been removed, though `Cow<'a, str>` remain. 
```

PR #153, opened 2021-03-19
Accept number of the form `1.`.

tomprince:


full-moon/tests/cases/pass/fractional-numbers/source.lua
```diff
@@ -1,4 +1,5 @@
 local num = 0.5
 local num2 = 0.5e5
 local num3 = .5
-local num4 = .5e5
\ No newline at end of file
+local num4 = .5e5
+local num4 = 1.
```
Kampfkarren, line 5:
```suggestion
local num5 = 1.
```

Kampfkarren (changes requested):
CHANGELOG as well, please.

full-moon/tests/cases/pass/fractional-numbers/source.lua
Kampfkarren: This'll need a reparse of the snap.

PR #154, opened 2021-03-20
Switch to using peg for lexing (and eventually parsing) instead of nom.

tomprince:
I was playing around (see #133) with converting the parser to use `nom`. 

I discovered that if you lean into using the contaminators:
- the code can get unwieldy (see for example `fn expression` in `ast::parsers` above
- because they are build of closures, lifetimes/mutability can get tricky (particularly if you want to reuse one parser in multiple places in a complex combinator expression.

So I decided to explore other options for parsing, and decided to experiment with [rust-peg](https://docs.rs/peg/). This PR is the result of converting (most of) the lexing to use rust-peg. It feels like this code is much easier to follow than the nom based version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Changed
+- Switched from using nom to peg for lexing.
+
 ## [0.10.0] - 2021-03-26
 ### Added
 - Added `block.iter_stmts_with_semicolon()` which returns an iterator over tuples containing the statement and an optional semicolon.
--- a/full-moon-derive/src/symbols.rs
+++ b/full-moon-derive/src/symbols.rs
@@ -10,8 +10,6 @@ use syn::{
     parse_macro_input, Ident, LitStr, Token,
 };
 
-const ALT_LIMIT: usize = 21;
-
 #[derive(Debug)]
 struct SymbolsInput {
     symbols: IndexMap<Ident, LitStr>,
@@ -65,33 +63,24 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let string: Vec<_> = symbols.values().collect();
     let ident: Vec<_> = symbols.keys().collect();
     let symbols: Vec<_> = symbols.iter().collect();
-    let splits = symbols.chunks(ALT_LIMIT).map(|symbols| {
-        let parsers = symbols.iter().map(|(symbol, string)| {
-            // Note this doesn't handle the case of keywords with digits
-            // which doesn't currently occur.
-            let matcher = if string
-                .value()
-                .chars()
-                .all(|char| char.is_ascii_alphabetic() || char == '_')
-            {
-                quote! {
-                    terminated(
-                        tag(#string),
-                        not(alt((alphanumeric1, tag("_")))),
-                    )
-                }
-            } else {
-                quote! {tag(#string)}
-            };
-            quote! {value(Symbol::#symbol, #matcher)}
-        });
-
-        quote! {
-            alt((#(
-                #parsers,
-            )*))
-        }
+    let (keywords, operators): (Vec<_>, Vec<_>) = symbols.iter().partition(|(_, string)| {
+        // Note this doesn't handle the case of keywords with digits
+        // which doesn't currently occur.
+        string
+            .value()
+            .chars()
+            .all(|char| char.is_ascii_alphabetic() || char == '_')
     });
+
+    let identifier_match: Vec<_> = keywords
+        .iter()
+        .map(|(symbol, string)| quote!(#string => Symbol::#symbol,))
+        .collect();
+
+    let operator_array: Vec<_> = operators
+        .iter()
+        .map(|(symbol, string)| quote!((Symbol::#symbol, #string)))
+        .collect();
 
     let output = quote! {
         /// A literal symbol, used for both words important to syntax (like while) and operators (like +)
@@ -126,20 +115,27 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
             }
         }
 
-        fn parse_symbol(code: &str) -> IResult<&str, Symbol> {
-            use ::nom::{
-                branch::alt,
-                bytes::complete::{tag},
-                character::complete::{alphanumeric1},
-                combinator::{not, value},
-                sequence::terminated,
-            };
-            let mut combinator = alt((
-                #(
-                    #splits,
-                )*
-            ));
-            combinator(code)
+        fn parse_keyword(identifier: &str) -> Option<Symbol> {
+            Some(match identifier {
+                #(#identifier_match)*
+                _ => return None,
+            })
+        }
+
+        trait ParseSymbol<'input> {
+            fn parse_symbol(self, pos: usize) -> peg::RuleResult<Symbol>;
+        }
+
+        impl<'input> ParseSymbol<'input> for &'input str {
+            fn parse_symbol(self: Self, pos: usize) -> peg::RuleResult<Symbol> {
+                for (symbol, string) in &[#(#operator_array,)*] {
+                    if self[pos..].starts_with(string) {
+                        return peg::RuleResult::Matched(pos + string.len(), *symbol);
+                    }
+                }
+
+                peg::RuleResult::Failed
+            }
         }
     };
 
--- a/full-moon/Cargo.toml
+++ b/full-moon/Cargo.toml
@@ -25,8 +25,8 @@ cfg-if = "1.0"
 derive_more = "0.99"
 full_moon_derive = { path = "../full-moon-derive", version = "=0.6.0" }
 paste = "0.1"
+peg = "0.6.3"
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
-nom = "6.1.2"
 
 [dev-dependencies]
 criterion = "0.2"
--- a/full-moon/Cargo.toml
+++ b/full-moon/Cargo.toml
@@ -25,7 +25,7 @@ cfg-if = "1.0"
 derive_more = "0.99"
 full_moon_derive = { path = "../full-moon-derive", version = "=0.6.0" }
 paste = "0.1"
-peg = "0.6.3"
+peg = "0.7.0"
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
 
 [dev-dependencies]
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -689,7 +689,7 @@ peg::parser! {
                 literal:v.1.into(),
                 quote_type: QuoteType::Brackets,
             }.into()}
-            / &multi_line_start() [_]* { TokenizerErrorType::UnclosedString.into() }
+            / &multi_line_start() [_]+ { TokenizerErrorType::UnclosedString.into() }
 
         rule escape()
             = "\\" [_]
@@ -728,7 +728,7 @@ peg::parser! {
             = "--" v:multi_line_block()
               { TokenType::MultiLineComment { blocks: v.0, comment: v.1.into() }.into() }
             / "--" multi_line_start() [_]* { TokenizerErrorType::UnclosedComment.into() }
-            / "--" comment:$((!['\r'|'\n'] [_])*)
+            / "--" comment:$(([^ '\r'|'\n'])*)
               { TokenType::SingleLineComment { comment: comment.into() }.into() }
 
         rule roblox()
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -1,17 +1,7 @@
 use crate::visitors::{Visit, VisitMut, Visitor, VisitorMut};
 
 use full_moon_derive::{symbols, Owned};
-use nom::{
-    branch::alt,
-    bytes::complete::{escaped, tag, tag_no_case, take_till, take_until, take_while, take_while1},
-    character::complete::{
-        alpha1, alphanumeric1, anychar, digit1, line_ending, none_of, one_of, space1,
-    },
-    combinator::{consumed, map, opt, recognize, success, value},
-    multi::{many0, many_till},
-    sequence::{delimited, pair, preceded, terminated},
-    IResult,
-};
+use peg;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, cmp::Ordering, fmt, str::FromStr};
@@ -95,6 +85,8 @@ pub enum TokenizerErrorType {
     UnclosedComment,
     /// An unclosed string was found
     UnclosedString,
+    /// An unexpected #! was found
+    UnexpectedShebang,
     /// An unexpected token was found
     UnexpectedToken(char),
     /// Symbol passed is not valid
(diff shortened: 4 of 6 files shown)


tomprince:
I think this is ready for review. Measuring the tokenizing benches using `iai`, I get between a 30-50% speed up over nom.

JohnnyMorganz:
Was messing around with this and seemed to have spotted an issue: attempting to parse `date.lua` in available in the benchmarks seems to cause an unexpected token error - the failure arises from a single line comment with nothing written afterwards, on line 3.
Adding this test to `test_rule_comment`:
```rs
test_rule!(
    comment("--"),
    TokenType::SingleLineComment { comment: "".into() }
);
```
gives out an `(internal full-moon error) Text overflow while giving unexpected token error`

full-moon/src/tokenizer.rs
```diff
@@ -955,345 +899,279 @@ impl std::error::Error for TokenizerError {}
...
-            assert_eq!($advancer($code), $result);
+    macro_rules! test_rule {
+        ($rule:ident($code:expr), $result:expr) => {
+            let code: &str = $code;
+            let result: RawToken = $result.into();
+
+            assert_eq!(
+                tokens::$rule(code)
+                    .map_err(|err| from_parser_error(code)(err).error)
+                    .and_then(|v| v),
```
Kampfkarren, line 970: Huh?

full-moon-derive/src/symbols.rs
```diff
@@ -126,20 +113,25 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        }
+
+        trait ParseSymbol<'input> {
+            fn parse_symbol(self, pos: usize) -> peg::RuleResult<Symbol>;
+        }
+        impl<'input> ParseSymbol<'input> for &'input str {
+            fn parse_symbol(self: Self, pos:usize) -> peg::RuleResult<Symbol> {
+                for (symbol, string) in &[#(#operator_array,)*] {
+                    if self[pos..].starts_with(string) {
+                        return peg::RuleResult::Matched(pos+string.len(), *symbol);
```
Kampfkarren, line 130: Rustfmt doesn't parse these, but they should still be aptly formatted.

```suggestion
                        return peg::RuleResult::Matched(pos + string.len(), *symbol);
```

full-moon-derive/src/symbols.rs
```diff
@@ -126,20 +113,25 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+                #(#identifier_match)*
+                _ => return None,
+            })
+        }
+
+        trait ParseSymbol<'input> {
+            fn parse_symbol(self, pos: usize) -> peg::RuleResult<Symbol>;
+        }
+        impl<'input> ParseSymbol<'input> for &'input str {
+            fn parse_symbol(self: Self, pos:usize) -> peg::RuleResult<Symbol> {
```
Kampfkarren, line 127:
```suggestion
            fn parse_symbol(self: Self, pos: usize) -> peg::RuleResult<Symbol> {
```

full-moon-derive/src/symbols.rs
```diff
@@ -126,20 +113,25 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+            Some(match identifier {
+                #(#identifier_match)*
+                _ => return None,
+            })
+        }
+
+        trait ParseSymbol<'input> {
+            fn parse_symbol(self, pos: usize) -> peg::RuleResult<Symbol>;
+        }
+        impl<'input> ParseSymbol<'input> for &'input str {
```
Kampfkarren, line 126:
```suggestion

        impl<'input> ParseSymbol<'input> for &'input str {
```

full-moon-derive/src/symbols.rs
```diff
@@ -65,33 +63,22 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        string
+            .value()
+            .chars()
+            .all(|char| char.is_ascii_alphabetic() || char == '_')
     });
+    let identifier_match: Vec<_> = keywords
+        .iter()
+        .map(|(symbol, string)| quote!(#string => Symbol::#symbol,))
+        .collect();
+    let operator_array: Vec<_> = operators
```
Kampfkarren, line 78:
```suggestion

    let operator_array: Vec<_> = operators
```

full-moon-derive/src/symbols.rs
```diff
@@ -65,33 +63,22 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+        .iter()
+        .map(|(symbol, string)| quote!((Symbol::#symbol , #string)))
```
Kampfkarren, line 80:
```suggestion
        .map(|(symbol, string)| quote!((Symbol::#symbol, #string)))
```

full-moon-derive/src/symbols.rs
```diff
@@ -126,20 +113,25 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+            fn parse_symbol(self, pos: usize) -> peg::RuleResult<Symbol>;
+        }
+        impl<'input> ParseSymbol<'input> for &'input str {
+            fn parse_symbol(self: Self, pos:usize) -> peg::RuleResult<Symbol> {
+                for (symbol, string) in &[#(#operator_array,)*] {
+                    if self[pos..].starts_with(string) {
+                        return peg::RuleResult::Matched(pos+string.len(), *symbol);
+                    }
+                }
+                peg::RuleResult::Failed
```
Kampfkarren, line 133:
```suggestion

                peg::RuleResult::Failed
```

full-moon/Cargo.toml
```diff
@@ -26,7 +26,7 @@ derive_more = "0.99"
 full_moon_derive = { path = "../full-moon-derive", version = "=0.5.0" }
 paste = "0.1"
 serde = { version = "1.0", features = ["derive", "rc"], optional = true }
-nom = "6.1.2"
+peg = "0.6.3"
```
Kampfkarren, line 29: Put this under `paste` for ABC order.

full-moon/src/tokenizer.rs
```diff
@@ -95,6 +85,8 @@ pub enum TokenizerErrorType {
     UnclosedComment,
     /// An unclosed string was found
     UnclosedString,
+    /// An unexpected #! was found
+    UnexpectedShebang,
```
Kampfkarren, line 89: Put this above UnclosedString, ABC order.

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,148 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+peg::parser! {
+    grammar tokens() for str {
+        use super::ParseSymbol;
+        use peg::ParseLiteral;
+        use super::StringLiteralQuoteType as QuoteType;
+        rule line_ending()
+            = "\n" / "\r\n"
+        rule space()
+            = [' '|'\t']
+        pub(super)
```
Kampfkarren, line 668: I know what pub(super) does, but what does it do in this context?

And for these, can you change it to `pub(super) rule whitespace() -> RawToken<'input>` so it's obvious what it's attached to?

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,148 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+        rule whitespace() -> RawToken<'input>
+            = chars:$( space()+ line_ending()? / line_ending() )
+              { TokenType::Whitespace {characters:chars.into() }.into()}
```
Kampfkarren, line 671:
```suggestion
              { TokenType::Whitespace { characters:chars.into() }.into() }
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,148 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+
+        rule multi_line_start() -> &'input str
+            = "[" block:$("="*) "[" {block}
```
Kampfkarren, line 674: Space all these so the rules are on their own pieces (as in, a newline after this).

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,148 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+            / number()
+            / string_literal()
+            / "#!" { TokenizerErrorType::UnexpectedShebang.into() }
+            / symbol()
+            / identifier()
+        pub(crate) rule tokens() -> Vec<(RawToken<'input>, usize)>
+            = shebang:(shebang:shebang() pos:position!() {(shebang,pos)})?
+              body:( token:token() pos:position!() {(token,pos)})*
+              {
+                  let mut body  = body;
```
Kampfkarren, line 781:
```suggestion
                  let mut body = body;
```

full-moon/src/tokenizer.rs
```diff
@@ -925,6 +805,7 @@ impl fmt::Display for TokenizerError {
             match &self.error {
                 TokenizerErrorType::UnclosedComment => "unclosed comment".to_string(),
                 TokenizerErrorType::UnclosedString => "unclosed string".to_string(),
+                TokenizerErrorType::UnexpectedShebang => "unexpected shebang".to_string(),
```
Kampfkarren, line 808: ABC order.

full-moon/src/tokenizer.rs
```diff
@@ -940,6 +821,69 @@ impl fmt::Display for TokenizerError {
...
+struct TokenCollector<'input> {
+    result: Vec<Token<'input>>,
+}
+
+// Collector
+impl<'input> TokenCollector<'input> {
+    fn new() -> Self {
+        Self { result: Vec::new() }
+    }
+    fn push(
```
Kampfkarren, line 843: Have you `cargo fmt`'d this? I'd think this would separate the new lines properly.

full-moon/src/tokenizer.rs
```diff
@@ -940,6 +821,69 @@ impl fmt::Display for TokenizerError {
...
+    ) -> Result<(), TokenizerError> {
+        match raw_token {
+            Ok(token_type) => {
+                self.result.push(Token {
+                    start_position,
+                    end_position,
+                    token_type,
+                });
+                Ok(())
+            }
```
Kampfkarren, line 857: Same here too, make sure you run `cargo fmt`!

Kampfkarren (changes requested):
Needs a formatting pass for sure, but are you confident this is actually better than nom? I know grammar, but this is a lot harder to read for me!

tomprince:
> [...] but are you confident this is actually better than nom? I know grammar, but this is a lot harder to read for me!

I don't have a very strong feeling either way. I can reasonably easily read either format. However, I find with nom, that I need to reason about the combinators, and basically convert to something like the peg format to understand what is going on.

There are some things that could be improved upstream, like https://github.com/kevinmehall/rust-peg/issues/240 and https://github.com/kevinmehall/rust-peg/commit/93a327b9f4eee0da0c91129af8495259447e0cc9 (which should be in the next release). It may also be that I've broken things down oddly, or that the grammar could use some more whitespace.

I'd also have a look at some the nom constructions in https://github.com/Kampfkarren/full-moon/pull/133, since I think the difference between nom and peg will be more pronounced for the ast as opposed to tokens (though I don't have a peg equivalent to compare to).

Kampfkarren:
I've been thinking about this, but I'm still not sure one way or the other, may ask some colleagues about it.

The nom parsers you've shown definitely do look ugly, and I don't like that it looks necessary to create entire structs around them, but I'm just a bit uncomfortable with such an un-Rust-like (whatever that means) DSL. Will hopefully get a solid opinion by the end of the week.

Kampfkarren:
I'm going to hesitantly approve on switching to peg. I'm still skeptical about introducing a DSL into the main code, but the nom parsers you showed definitely don't look less complex.

full-moon/src/tokenizer.rs
tomprince: `Unexpected..` come after `Unclosed...`, though I could sort on the second word.

full-moon/src/tokenizer.rs
tomprince: > I know what pub(super) does, but what does it do in this context?

`peg::parser!` creates a module, so `pub(super)` exposes the rule to the containing (i.e. `tokenizer`) module as a function.

full-moon/src/tokenizer.rs
tomprince: The rules internally return a `RawToken` which is (at least currently) as `Result` with an error of `TokenizerErrorType`, but peg wraps the error into its own result with its own error type. The `map_err` maps the peg error to ours, resulting in a `Result<Result<T, TokenizerErrorType>, TokenizerErrorType>`. The `and_then` flattens this to `Result<T, TokenizerErrorType>` (I'd use [`.flatten`](https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten) if it was stable, which has exactly this implementation).

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Added
+
+### Changed
+- Switched from using nom to peg for lexing.
```
Kampfkarren, line 11: Remove Added and the rest and just keep Changed until we need it.

full-moon-derive/src/symbols.rs
```diff
@@ -65,33 +63,23 @@ pub fn parse(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
-        }
+    let (keywords, operators): (Vec<_>, Vec<_>) = symbols.iter().partition(|(_, string)| {
+        // Note this doesn't handle the case of keywords with digits
+        // which doesn't currently occur.
+        string
+            .value()
+            .chars()
+            .all(|char| char.is_ascii_alphabetic() || char == '_')
     });
+    let identifier_match: Vec<_> = keywords
```
Kampfkarren, line 74: New line before this.

full-moon/src/tokenizer.rs
Kampfkarren: I just misread it, oops.

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+        use peg::ParseLiteral;
+        use super::StringLiteralQuoteType as QuoteType;
+        rule line_ending()
+            = "\n" / "\r\n"
+        rule space()
+            = [' '|'\t']
+
+        pub(super) rule whitespace() -> RawToken<'input>
+            = chars:$( space()+ line_ending()? / line_ending() )
+              { TokenType::Whitespace { characters:chars.into() }.into() }
```
Kampfkarren, line 671:
```suggestion
              { TokenType::Whitespace { characters: chars.into() }.into() }
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+        use super::ParseSymbol;
+        use peg::ParseLiteral;
+        use super::StringLiteralQuoteType as QuoteType;
+        rule line_ending()
+            = "\n" / "\r\n"
+        rule space()
+            = [' '|'\t']
+
+        pub(super) rule whitespace() -> RawToken<'input>
+            = chars:$( space()+ line_ending()? / line_ending() )
```
Kampfkarren, line 670:
```suggestion
            = chars:$( space() + line_ending()? / line_ending() )
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
-}
+peg::parser! {
+    grammar tokens() for str {
+        use super::ParseSymbol;
+        use peg::ParseLiteral;
+        use super::StringLiteralQuoteType as QuoteType;
+        rule line_ending()
+            = "\n" / "\r\n"
+        rule space()
+            = [' '|'\t']
```
Kampfkarren, line 667:
```suggestion
            = [' ' | '\t']
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
-            },
-        })),
-        Err(_) => Ok(None),
-    }
-}
+peg::parser! {
+    grammar tokens() for str {
+        use super::ParseSymbol;
+        use peg::ParseLiteral;
+        use super::StringLiteralQuoteType as QuoteType;
```
Kampfkarren, line 663:
```suggestion
        use super::StringLiteralQuoteType as QuoteType;
        
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+        rule multi_line_block() -> (usize, &'input str)
+            = block:multi_line_start()
+              content:$((!multi_line_end(block) [_])*)
+              multi_line_end(block)
+              { (block.len(), content) }
+
+        rule multi_line_quote() -> RawToken<'input>
+            = v:multi_line_block() { TokenType::StringLiteral {
+                multi_line: Some(v.0),
+                literal:v.1.into(),
```
Kampfkarren, line 688:
```suggestion
                literal: v.1.into(),
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+                quote_type: QuoteType::Brackets,
+            }.into()}
```
Kampfkarren, line 690:
```suggestion
            }.into() }
```

full-moon/src/tokenizer.rs
```diff
@@ -650,260 +642,153 @@ impl<'a> fmt::Display for StringLiteralQuoteType {
...
+            / symbol()
+            / identifier()
+
+        pub(crate) rule tokens() -> Vec<(RawToken<'input>, usize)>
+            = shebang:(shebang:shebang() pos:position!() {(shebang,pos)})?
+              body:( token:token() pos:position!() {(token,pos)})*
+              {
+                  let mut body = body;
+                  if let Some(shebang) = shebang {
+                      body.insert(0, shebang)
```
Kampfkarren, line 788:
```suggestion
                      body.insert(0, shebang);
```

full-moon/src/tokenizer.rs
```diff
@@ -940,6 +826,69 @@ impl fmt::Display for TokenizerError {
...
+        raw_token: RawToken<'input>,
+        end_position: Position,
+    ) -> Result<(), TokenizerError> {
+        match raw_token {
+            Ok(token_type) => {
+                self.result.push(Token {
+                    start_position,
+                    end_position,
+                    token_type,
+                });
```
Kampfkarren, line 860:
```suggestion
                });
                
```

full-moon/src/tokenizer.rs
```diff
@@ -940,6 +826,69 @@ impl fmt::Display for TokenizerError {
...
+            }),
+        }
+    }
+    fn finish(mut self, eof_position: Position) -> Vec<Token<'input>> {
+        self.result.push(Token {
+            start_position: eof_position,
+            end_position: eof_position,
+            token_type: TokenType::Eof,
+        });
+        self.result
```
Kampfkarren, line 875:
```suggestion

        self.result
```

full-moon/src/tokenizer.rs
Kampfkarren: Ah.

Kampfkarren (changes requested):
I, really am not quite sure how to approach this. The tests pass, which should be good enough for me. Here's some pedantic style quirks in the meantime. 😛

full-moon/src/tokenizer.rs
tomprince: This is a trailing operator for 1 or more repeats, not addition.

Kampfkarren: Oops, good catch.

PR #164, opened 2021-04-17
Deny clippy warnings in CI and fix issues

JohnnyMorganz:
Denies warnings in CI for clippy, so that the workflow fails.
Fixes currently outstanding warnings

Note, I had to allow `unnecessary_wraps` in the `tokenizer.rs` file due to all the `advance_XXXX()` methods, as clippy currently triggers a false positive for these cases (https://github.com/rust-lang/rust-clippy/issues/6613, and maybe https://github.com/rust-lang/rust-clippy/issues/6517?)

### DIFF
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -1,7 +1,6 @@
 use crate::visitors::{Visit, VisitMut, Visitor, VisitorMut};
 
 use full_moon_derive::{symbols, Owned};
-use peg;
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 use std::{borrow::Cow, cmp::Ordering, fmt, str::FromStr};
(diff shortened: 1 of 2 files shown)


full-moon/src/tokenizer.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::unnecessary_wraps)] // Clippy currently triggers false positives wrt. Advancement: https://github.com/rust-lang/rust-clippy/issues/6613
```
Kampfkarren, line 1: Where was this popping up? Only in a few places, or everywhere? If it's not in a lot of places, I'd rather filter those places specifically.

JohnnyMorganz: It was for all the following:
- `advance_identifier`
- `advance_shebang`
- `advance_symbol`
- `advance_whitespace`
The only one it was fine for was `advance_quote`

JohnnyMorganz: Following the merge of #154, this is no longer necessary. It has now been removed

Kampfkarren:
Thanks!

PR #166, opened 2021-04-17
Fix the tokens order for Node trait

### LINKED ISSUE #161: Node `.tokens()` is not correctly ordered for ContainedSpans
If you have an expression node contained in parentheses, e.g. `(foo)` in `local x = (foo)`, and then call `.tokens()` on it, then it is expected that the output is in the order of `LeftParens`, `foo`, `RightParens`.

However, it is in the order `LeftParens`, `RightParens`, `foo`. 

This issue also affects `Node.surrounding_trivia()`, as the surrounding trivia is expected to be `(Leading Trivia of LeftParens, Trailing Trivia of RightParens)`, however since it is not correctly ordered, what is actually returned is `(Leading Trivia of LeftParens, Trailing Trivia of foo)`.

JohnnyMorganz:
Fixes #161 
Also reorders struct fields in places where the current order was incorrect

Currently only implemented for `TableConstructor`, to see how it would work. Also added a test case for it.

### DIFF
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -2302,7 +2302,11 @@ pub(crate) fn extract_token_references(mut tokens: Vec<Token>) -> Vec<TokenRefer
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{parse, print, tokenizer::tokens, visitors::VisitorMut};
+    use crate::{
+        parse, print,
+        tokenizer::tokens,
+        visitors::{Visitor, VisitorMut},
+    };
 
     #[test]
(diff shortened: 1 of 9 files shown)


full-moon/src/ast/mod.rs
```diff
@@ -260,6 +261,31 @@ impl Default for TableConstructor<'_> {
     }
 }
 
+impl Sealed for TableConstructor<'_> {}
+impl<'a> NodeTrait<'a> for TableConstructor<'a> {
```
Kampfkarren, line 265: We have a separate file for these implementations instead of mod.rs, which is already pretty beefy.

PR #167, opened 2021-04-20
Add variadic type support and restrict where some types are allowed

### LINKED ISSUE #91: Typed Lua - Tuple syntax is not supported
```lua
function foo(): (number, number)
```

...produces a syntax error.

### LINKED ISSUE #162: Add support for Luau variadic types
https://roblox.github.io/luau/typecheck#variadic-types

JohnnyMorganz:
Closes #162 

Tuple types are only permitted as the return type of a function (both for function declarations, or for callback type annotations)
Variadic Types are permitted only as a parameter in a callback type annotation, as a return type, or inside of a tuple acting as a return type.
This PR implements these restrictions, and fails parsing otherwise - fail cases have been added to the tests
Fixes #91

### DIFF
--- a/full-moon/src/ast/parser_util.rs
+++ b/full-moon/src/ast/parser_util.rs
@@ -179,6 +179,12 @@ macro_rules! keep_going {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! define_roblox_parser {
+    ($parser:ident, $node:ty, $mock_ty:ty, |_, $state:ident| $body:expr) => {
+        define_parser! {$parser, $node, |_, mut $state: ParserState<'a, 'b>| $body}
+    };
+    ($parser:ident, $node:ty, $mock_ty:ty, |$self:ident, $state:ident| $body:expr) => {
+        define_parser! {$parser, $node, |$self:&$parser, mut $state: ParserState<'a, 'b>| $body}
+    };
     ($parser:ident, $node:ty, $mock_ty:ty, $body:expr) => {
         cfg_if::cfg_if! {
             if #[cfg(feature = "roblox")] {
--- a/full-moon/src/ast/parsers.rs
+++ b/full-moon/src/ast/parsers.rs
@@ -341,7 +341,7 @@ define_roblox_parser!(
         let (state, assertion_op) = ParseSymbol(Symbol::TwoColons).parse(state)?;
         let (state, cast_to) = expect!(
             state,
-            ParseTypeInfo.parse(state),
+            ParseTypeInfo(TypeInfoContext::None).parse(state),
             "expected type in type assertion"
         );
 
@@ -792,7 +792,7 @@ define_parser!(ParseFunctionBody, FunctionBody<'a>, |_, state| {
 
                 // Parse ellipse type if in Luau
                 if cfg!(feature = "roblox") {
-                    if let Ok((new_state, type_specifier)) = ParseTypeSpecifier.parse(state) {
+                    if let Ok((new_state, type_specifier)) = ParseTypeSpecifier(TypeInfoContext::None).parse(state) {
                         state = new_state;
                         type_specifiers.push(Some(type_specifier));
                     } else {
@@ -807,7 +807,7 @@ define_parser!(ParseFunctionBody, FunctionBody<'a>, |_, state| {
 
         // Parse ellipse type if in Luau
         if cfg!(feature = "roblox") {
-            if let Ok((new_state, type_specifier)) = ParseTypeSpecifier.parse(state) {
+            if let Ok((new_state, type_specifier)) = ParseTypeSpecifier(TypeInfoContext::None).parse(state) {
                 state = new_state;
                 type_specifiers.push(Some(type_specifier));
             } else {
@@ -823,7 +823,7 @@ define_parser!(ParseFunctionBody, FunctionBody<'a>, |_, state| {
     );
 
     #[cfg_attr(not(feature = "roblox"), allow(unused_variables))]
-    let (state, return_type) = if let Ok((state, return_type)) = ParseTypeSpecifier.parse(state) {
+    let (state, return_type) = if let Ok((state, return_type)) = ParseTypeSpecifier(TypeInfoContext::ReturnType).parse(state) {
         (state, Some(return_type))
     } else {
         (state, None)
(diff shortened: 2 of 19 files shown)


JohnnyMorganz:
Everything seems to work fine, apart from a variadic type on the LHS of a Union/Intersection type (e.g. `...string | number`).
This incorrectly gets parsed as
```
Variadic
    ellipse: ...
    type_info:
        Union:
            lhs: string
            pipe: |
            rhs: number
```
when it should be parsed as
```
Union
    lhs: Variadic:
        ellipse: ...
        type_info: string
    pipe: |
    right: number
```
Need to find a way to fix this

full-moon/tests/roblox_cases/pass/types_variadic/source.lua
```diff
@@ -0,0 +1,4 @@
+type Foo = (...number) -> ()
+
+function bar(...: number)
```
matthargett, line 3: I would add a few more test cases here:
```lua
type Foo = (...number) -> (string, ...number) -> ...any
type Baz = (string, ...Foo) -> ...Foo

local Boo = {}
function Boo:f(name: string, ...: number): () -> (...number) -> ()
  return function()
    return function(_x: string, ...: Foo) end
  end
end
```

Kampfkarren:
Have you figured out your problem? Is this ready for merge?

JohnnyMorganz:
> Have you figured out your problem? Is this ready for merge?

Yeah I think so, but it's probably a good idea for you to double check to see if you're happy with how the parsing is working

full-moon/src/ast/parser_util.rs
```diff
@@ -179,6 +179,12 @@ macro_rules! keep_going {
 #[doc(hidden)]
 #[macro_export]
 macro_rules! define_roblox_parser {
+    ($parser:ident, $node:ty, $mock_ty:ty, |_, $state:ident| $body:expr) => {
```
Kampfkarren, line 182: Do we need this? `_` is a valid identity.

full-moon/src/ast/parser_util.rs
```diff
@@ -179,6 +179,12 @@ macro_rules! keep_going {
...
+        define_roblox_parser! {$parser, $node, $mock_ty, |_, mut $state: ParserState<'a, 'b>| $body}
+    };
+    ($parser:ident, $node:ty, $mock_ty:ty, |$self:ident, $state:ident| $body:expr) => {
+        define_roblox_parser! {$parser, $node, $mock_ty, |$self:&$parser, mut $state: ParserState<'a, 'b>| $body}
```
Kampfkarren, line 186: I'm confused, how does this work? This doesn't appear to be a valid definition of `define_roblox_parser`, all the current definitions use `()`.

full-moon/src/ast/parsers.rs
```diff
@@ -789,11 +789,31 @@ define_parser!(ParseFunctionBody, FunctionBody<'a>, |_, state| {
                 parameters.push(last_parameter);
 
                 parameters.push(Pair::new(Parameter::Ellipse(ellipse), None));
+
+                // Parse ellipse type if in Luau
+                if cfg!(feature = "roblox") {
```
Kampfkarren, line 794: This needs to be `cfg_if`.

This is going to compile to `if false`, but the code inside it will still attempt to compile. `ParserTypeSpecifier` and friends don't (or rather, shouldn't), exist without the `roblox` feature flag, so this won't compile.

full-moon/src/ast/parsers.rs
```diff
@@ -789,11 +789,31 @@ define_parser!(ParseFunctionBody, FunctionBody<'a>, |_, state| {
...
+                    }
+                }
             }
         }
     } else if let Ok((new_state, ellipse)) = ParseSymbol(Symbol::Ellipse).parse(state) {
         state = new_state;
         parameters.push(Pair::new(Parameter::Ellipse(ellipse), None));
+
+        // Parse ellipse type if in Luau
+        if cfg!(feature = "roblox") {
```
Kampfkarren, line 809: Same here.

full-moon/src/ast/parsers.rs
```diff
@@ -1349,30 +1387,57 @@ cfg_if::cfg_if! {
...
+                            (
+                                state,
+                                TypeInfo::Tuple {
+                                    parentheses: ContainedSpan::new(start_parenthese, end_parenthese),
+                                    types,
+                                },
+                            )
+                        }
+                    }
+                    _ => {
```
Kampfkarren, line 1418: Just use `if let`, clippy should've caught this.

Kampfkarren (changes requested):
The parsing results looks fine, ultimately you're the consumer of these, so if you think they're fine to work with then it's alright with me.

full-moon/src/ast/parsers.rs
Kampfkarren: For whatever reasons the tests say this is fine, which means `ParseTypeSpecifier` is for whatever reason on non-roblox builds.

This should still probably be cfg_if.

full-moon/src/ast/parser_util.rs
JohnnyMorganz: I mainly copied this off how `define_parser!` macro works, it's the same layout there. AIUI, since this is a macro matching on a specific token, it's different

full-moon/src/ast/parsers.rs
JohnnyMorganz: I did not change this from how it was previously done, so yeah ParseTypeSpecifier was present in non-roblox builds (I think this is what the `define_roblox_parser!` macro is for? It falls back to a noop which will never run, parsing a TokenReference in non roblox builds)

I can change this to cfg_if though

full-moon/src/ast/parser_util.rs
JohnnyMorganz: I'm not entirely sure what you're referring to here, but I just copied and modified the layout of the define_parser macro to fit here

Kampfkarren: I'm referring to `define_roblox_parser! {}` using braces instead of parentheses, when I don't see that being a valid macro format.

full-moon/src/ast/parsers.rs
Kampfkarren: Oh, fair enough, I forgot.

full-moon/src/ast/parser_util.rs
JohnnyMorganz: It is a valid format. The reason its using `{}` instead of `()` is that the macro expands to items.
The alternative could be to leave it as `()` and add a semicolon to the end, but this way works fine and is valid.
I quickly googled and its mentioned here; this document is old but it still seems to be valid: http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/macros.html#syntactic-requirements

full-moon/src/ast/parser_util.rs
JohnnyMorganz: Just tested this, and you're right - I'll remove this definition. Not sure if you want to do the same for `define_parser!`?

Kampfkarren: Probably should.

full-moon/src/ast/parser_util.rs
JohnnyMorganz: Just looking at the change, it seems rustfmt formats it into something quite ugly...
Should we add a skip attribute, revert to the original, or just leave it?

PR #177, opened 2021-05-21
Fix a single type wrapped in parentheses failing parsing

### LINKED ISSUE #173: error parsing while creating ast: unexpected token `?` in nil-able function type
```
error parsing: error occurred while creating ast: unexpected token `?`. (starting from line 106, character 85 and ending on line 106, character 86)
additional information: expected `->` when parsing function type
```

for this code:
```lua
function TypeInfo.new(
        schema: Schema,
        -- NOTE: this experimental optional second 
        getFieldDefFn: ((Schema, GType, FieldNode) -> GField<any, any>?)?
)
end
````

### LINKED ISSUE #174: error occurred while creating ast: unexpected token `|` in type that is a function returning type or type
```
error parsing: error occurred while creating ast: unexpected token `|`. (starting from line 596, character 34 and ending on line 596, character 35)
additional information: expected `->` when parsing function type
```

for this code:
```lua
export type Thunk<T> = (() -> T) | T
```

### LINKED ISSUE #175: error occurred while creating ast: unexpected token `:` with nil-able function types
```
error parsing: error occurred while creating ast: unexpected token `:`. (starting from line 2, character 12 and ending on line 2, character 13)
additional information: expected `}` to match `{`
```

for this code:
```
export type PromiseLike<T> = {
    andThen: (
                ((T) -> T)? | (PromiseLike<T>)?, -- resolve
                ((any) -> () | PromiseLike<T>)? -- reject
        ) -> PromiseLike<T>
}
```

### LINKED ISSUE #176: support type intersection syntax
```
function GError.new(
        message: string,
        originalError: (Error & { extensions: any? }), -- new syntax
        extensions: { [string]: any }? 
): GError
```

Simple solution: Anywhere a `|` can be used in a typedef can also be an `&`

JohnnyMorganz:
Regression caused by #167, where we accidently disallowed a single type wrapped around in parentheses in places which were not a return type.

This is valid syntax, and is necessary in cases where we want to declare the type as optional, such as:
```lua
type Foo = (Bar & Baz)?
```

Fixes #173 
Fixes #174 
Fixes #175 
Fixes #176

### DIFF
--- a/full-moon/src/ast/parsers.rs
+++ b/full-moon/src/ast/parsers.rs
@@ -1409,9 +1409,13 @@ cfg_if::cfg_if! {
                     match this.0 {
                         TypeInfoContext::ReturnType => {},
                         _ => {
-                            if types.len() != 1 {
+                            let num_types_present = types.len();
+                            if num_types_present != 1 {
                                 return Err(InternalAstError::UnexpectedToken {
-                                    token: state.peek().clone(), // TODO: this is pointing to the wrong thing
+                                    token: match num_types_present {
+                                        0 => end_parenthese,
+                                        _ => types.pairs().next().unwrap().punctuation().unwrap().clone(), // A comma after the first type must exist if > 1 types present
+                                    },
                                     additional: Some("tuples are only permitted as a return type"),
                                 });
                             };
--- a/full-moon/src/ast/parsers.rs
+++ b/full-moon/src/ast/parsers.rs
@@ -1387,45 +1387,12 @@ cfg_if::cfg_if! {
                     "expected `)` to match `(`"
                 );
 
-                if let TypeInfoContext::ReturnType = this.0 {
-                    // Tuples are only permitted as the return type of a function
-                    if let Ok((state, arrow)) = ParseSymbol(Symbol::ThinArrow).parse(state) {
-                        let (state, return_value) = expect!(
-                            state,
-                            ParseTypeInfo(TypeInfoContext::ReturnType).parse(state),
-                            "expected return type after `->`"
-                        );
-                        (
-                            state,
-                            TypeInfo::Callback {
-                                arguments: types,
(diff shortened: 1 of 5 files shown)


Kampfkarren:
That clippy is unrelated to you, if you don't get to it I'll just push it later tomorrow.

full-moon/src/ast/parsers.rs
```diff
@@ -1435,6 +1402,33 @@ cfg_if::cfg_if! {
...
+
+                    match this.0 {
+                        TypeInfoContext::ReturnType => {},
+                        _ => {
+                            let num_types_present = types.len();
+                            if num_types_present != 1 {
+                                return Err(InternalAstError::UnexpectedToken {
+                                    token: match num_types_present {
+                                        0 => end_parenthese,
+                                        _ => types.pairs().next().unwrap().punctuation().unwrap().clone(), // A comma after the first type must exist if > 1 types present
```
Kampfkarren, line 1417: What the heck is this doing?

Kampfkarren: Can you use `expect` both to document what's actually going on and in case this fails, we'll know what it is?

JohnnyMorganz: This is a weird one. If we reach this match, it means there was either 0 or 1+ types inside the parentheses, which are only possible as a function return type. I had it simpler before the last commit, but I had to do this so that the error would point to the right token, in this case the comma after the first type.

I don't think `except` is possible here - we have already parsed the tokens before we reached here, now we are checking what we have parsed to see if it's valid. Im busy for the next 2 weeks, but I can look into this further to make it more understandable. (Feel free to also change it yourself if you want)

Kampfkarren: `expect` as in, instead of `unwrap` to document what went wrong if they failed.

PR #184, opened 2021-06-19
Fix parse issue with complex function type args

JohnnyMorganz:
They begin with an identifier, but we assumed it would've been a name, but it might not be


Kampfkarren:
Thanks!

PR #196, opened 2021-10-02
Handle UTF8 BOM

### LINKED ISSUE #117: Parsing fails on files with BOM
I just tried parsing a few files with Selene that had a UTF8 BOM and it failed to parse them, the error was pretty generic, should full-moon work with BOM or should the error point out BOM and tell users to remove it?

zhiburt:
closes #117

Curiosity: I was wondering if original Lua intepretator correctly handles this?

From 5 mines of wondering I would say it doesn't?
Because it doesn't skip chars actually?
Am I wrong here?

https://github.com/lua/lua/blob/439e45a2f69549b674d6a6e2023e8debfa00a2b8/lauxlib.c#L742-L753

### DIFF
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -781,7 +781,7 @@ peg::parser! {
                       body.insert(0, shebang)
                   }
                   if let Some(bom) = bom {
-                    body.insert(0, bom)
+                      body.insert(0, bom)
                   }
                   body
               }
--- a/full-moon/src/tokenizer.rs
+++ b/full-moon/src/tokenizer.rs
@@ -702,6 +702,10 @@ peg::parser! {
             = line:$("#!" (!line_ending() [_])* line_ending())
               {TokenType::Shebang{line:line.into()}.into()}
 
+        pub(super) rule utf8_bom() -> RawToken
+            = chars:$("\u{feff}") // A BOM sequence in UTF8 "\xEF\xBB\xBF"
+            { TokenType::Whitespace { characters:chars.into() }.into() }
+
         pub(super) rule identifier() -> RawToken
             = id:$(['_'|'a'..='z'|'A'..='Z'] ['_'|'a'..='z'|'A'..='Z'|'0'..='9']*)
               { match parse_keyword(id) {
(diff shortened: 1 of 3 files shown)


JohnnyMorganz:
As full-moon is a lossless parser, should the BOM be explicitly handled as a whitespace/trivia token and left in the AST, rather than discarded completely?
Is it important for the BOM to be available, I don't really know

Kampfkarren:
> As full-moon is a lossless parser, should the BOM be explicitly handled as a whitespace/trivia token and left in the AST, rather than discarded completely?

Yeah, it should.

This is all assuming Lua 5.1 doesn't error on this as well, which I should double check.

zhiburt:
So what you are saying it's better to be handled as a token and then be present in AST?

Kampfkarren:
We have existing whitespace tokens, and so would probably be best to put them in.

zhiburt:
Aaa but it's not whitespace exactly?

LuaJIT seems skips it as well

https://github.com/LuaJIT/LuaJIT/blob/1b7171c339a8d33cb1fd332e31787ebc23266f10/src/lj_lex.c#L415-L420

JohnnyMorganz:
As a loss-less parser, running `full_moon::print(full_moon::parse(&code)?)` should preserve all content, including trivia/whitespace, and in this case the BOM.

I'm not sure if it should just be added to the whitespace tokens, or if another token should be made, like Shebang?

Kampfkarren:
A BOM token seems like a fine idea then, I hope that's already marked as non_exhaustive (I think it is).

zhiburt:
I did reuse `Whitespace`

full-moon/src/tokenizer.rs
```diff
@@ -768,13 +772,17 @@ peg::parser! {
...
+            = bom:(bom:utf8_bom() pos:position!() {(bom,pos)})?
+              shebang:(shebang:shebang() pos:position!() {(shebang,pos)})?
               body:( token:token() pos:position!() {(token,pos)})*
               {
                   let mut body = body;
                   if let Some(shebang) = shebang {
                       body.insert(0, shebang)
                   }
+                  if let Some(bom) = bom {
+                    body.insert(0, bom)
```
Kampfkarren, line 784: Can you run Cargo fmt over this? This doesn't look right

If Cargo fmt can't pick this up (since it's in a macro), can you keep this consistent with existing indents?

Kampfkarren: Also, this should probably be before the shebang insertion, right?

Kampfkarren (changes requested):

full-moon/src/tokenizer.rs
zhiburt: > This doesn't look right

I see

> Also, this should probably be before the shebang insertion, right?

It doesn't because it must be BEFORE shebang

Kampfkarren: Oh, right, I forgot this was insert.

Kampfkarren:
Last thing, can you add this in CHANGELOG? This would be under "Added", under "Unreleased".

Kampfkarren:
Thanks!

PR #203, opened 2021-10-21
Fix returned function types with named parameters

### LINKED ISSUE #201: Luau: Named function parameters in function return value
Currently, specifying a function's return type with named type arguments results in a parse error.
image
In Roblox Studio, no errors are shown.

Putting parentheses around the return type or removing the names gets rid of the error.
image

This error doesn't happen anywhere else. 
image

JohnnyMorganz:
Fixes #201

### DIFF
--- a/full-moon/tests/roblox_cases/pass/named_function_arg_types/source.lua
+++ b/full-moon/tests/roblox_cases/pass/named_function_arg_types/source.lua
@@ -6,3 +6,6 @@ end
 
 local function bar(x: (number)?): (baz: string) -> string
 end
+
+local function bar(x: (number)?): (baz: string) -> ((names: Array<string>) -> ...any)
+end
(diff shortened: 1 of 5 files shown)


JohnnyMorganz:
> Code looks like it would handle it, but maybe add another test that has a second level of typed function return with named arg?

Do you mind giving an example? Not sure exactly what you mean

Kampfkarren:
I also would like to know what you mean, for more test coverage @matthargett

full-moon/tests/roblox_cases/pass/named_function_arg_types/source.lua
```diff
@@ -2,4 +2,7 @@ type MyCallbackType = (cost: number, name: string) -> string
 
 local cb: (amount: number) -> number
 local function foo(cb: (name: string) -> ())
+end
+
```
matthargett, line 6:
```suggestion

local function bar(x: (number)?): (baz: string) -> ((names: Array<string>) -> ...any)
end

```

matthargett (approved):

PR #207, opened 2021-10-30
Added if expression support

JohnnyMorganz:
Adds support for if expressions (with arbitrary number of `elseif` expressions present).
Available under the roblox feature flag

### DIFF
--- a/full-moon/tests/roblox_cases/pass/if_expression/source.lua
+++ b/full-moon/tests/roblox_cases/pass/if_expression/source.lua
@@ -1,3 +1,6 @@
 local x = if foo then foo.x else 5
 local y = (if x then x.indices else create()):update(if shouldUpdate then information else defaults)
-local y = (if bar then foo.y else 5) :: number
+local z = (if bar then foo.y else 5) :: number
+
+local a = if foo then foo.x elseif bar then bar.x else 5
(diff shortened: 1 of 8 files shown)


Kampfkarren:
Tests failing

full-moon/tests/roblox_cases/pass/if_expression/source.lua
```diff
@@ -0,0 +1,6 @@
+local x = if foo then foo.x else 5
+local y = (if x then x.indices else create()):update(if shouldUpdate then information else defaults)
+local z = (if bar then foo.y else 5) :: number
+
+local a = if foo then foo.x elseif bar then bar.x else 5
```
matthargett, line 5: a nuance to capture:
```suggestion
local a = if foo then (foo.x :: number) elseif bar then bar.x()() else 5
```

matthargett (approved):
Minor suggestion to add a more complex test case, but otherwise looks great!

full-moon/tests/roblox_cases/pass/if_expression/source.lua
Kampfkarren: We should add this as another test case, instead of updating the existing one, but yeah!

matthargett (approved):

PR #213, opened 2021-12-08
Add support for plugins to change parsing, printing, and visiting behaviors

### LINKED ISSUE #152: Allow custom parsers to be created, separate roblox and 5.2 features into their own crates
This is extremely tired rambling, so bear with me. 

I think it's uncomfortable that full-moon must undergo a major version change nearly every time Luau (Roblox's Lua) changes something about its syntax, as it forces ALL consumers to step up (or we have to maintain two versions). 

I think, if possible, we should be able to create extensions for full-moon (not dylibs) as their own crates, such as full-moon-luau, which would be able to provide parsers to full-moon so that only it would need to undergo a major version change. 

Do this with 5.2 too, and support both (perhaps provide this with a feature flag, as Lua 5.2 won't change). Base full-moon would be 5.1. 

Solves #47?

Kampfkarren:
Won't close #152 until we actually *do* make those plugins (especially the Luau one, which I don't really want to maintain, at least not alone).

Will write extensive documentation later.

Point is every node now has a `P: Plugin` generic (defaulting to `DefaultPlugin`) which links to structs that implement `PluginMod` for every node (`AssignmentMod: PluginMod<Assignment<Self>>`), which contains the information for what values to put in each node, as well as display, and when I finish it, visiting. Plugins will also likely be where custom parsing is handled. You will only be able to have one plugin at a time.

This is all designed to, ideally (though not tested), **be completely zero-cost if no plugin is used**. `DefaultPlugin` has `()` as its NodeInfo for structs, and `Never` (an empty enum, ideally `!` in the future?) for enums. `()` is a [ZST](https://doc.rust-lang.org/nomicon/exotic-sizes.html#zero-sized-types-zsts), meaning it will not take up any additional space in the struct, and `Never` means that the branch should functionally not exist in the eyes of the compiler, since it is uncreatable in safe Rust.

Creating plugins is a bit obtuse, but shouldn't be a big deal since the number of consumers will be very small, and probably kept in close contact with me.

- [x] Custom struct/enum information
- [x] Custom parsers
- [ ] Custom visitors
- [ ] Port Luau
- [ ] Port Lua 5.2
- [ ] Documentation
- [ ] Proc macro to supplement default associated types


rzvxa:
Hello @Kampfkarren, Sorry for pinging you on this old PR, But I couldn't help but wonder why this got closed. Were there any challenges with the current structure of the parser that makes it impossible to decouple stuff?

I'm working on a gradually typed language that compiles to Lua and can target all runtimes. In some sense, it is similar to TypeScript, But it is not really a superset; While most things in the syntax are close to Lua, and Luau, The language comes with its own quirks such as nil safety, pattern matching, if expression among other things.

Since full-moon has support for all of the target runtimes, It would be nice if it was possible to have an extendable system that would allow having multiple AST types simultaneously.

I would happily work on it with you, Was there ever a plan to also make AST extendable? Either way, I'm probably going to base my compiler's parser on this project and I prefer to contribute it back to this project. There are going to be some generalizations to support a higher-level syntax(such as the macro system and other new features). I know at least part of it could be welcomed here with this PR.

If we want to decouple different language features from the parser we should consider also modularizing the AST itself, Which in return needs to have extendable `Stmt`, `Expression`, etc. It would introduce some sort of dynamic dispatching into the mix which I'm not sure is the most desirable thing. On the other hand, if we can't make the AST extendable without introducing too much overhead I need to hard fork it for extension and with each Luau update or new Lua release or on the off chance that this language gets some actual users in the wild, we all get to update our versions respectively.

I haven't read the source thoroughly so I need more time before I can have a solid understanding of the project structure, Have you considered extending plugins to AST itself? If it was part of the plan, what was your idea for tackling it? I would appreciate it if you let me know.

PR #217, opened 2022-01-13
Clipped and ready

Rerumu:
Addresses a few Clippy lints regarding idiomatic iterators, asserts, and if statement ordering.


Kampfkarren:
Thanks!

PR #219, opened 2022-01-19
Singleton types

matthargett:
Adds support for Luau's singleton types feature, merged Luau RFC [here](https://github.com/Roblox/luau/blob/master/rfcs/syntax-singleton-types.md).

Examples:
```lua
type Element = { ["$$typeof"]: number }
type which = "Query" | "Mutation" | "Subscription"
type NetworkError = Error & { wasHandled: true }
```

Fixes an issue where invalid code would parse:
```lua
type nil<boolean>
```

Tweaked some of the roblox_cases tests so they pass strict mode analysis with OSS Luau, allowing for a more direct, easy cross-check.


full-moon/src/ast/parser_util.rs
```diff
@@ -76,7 +76,7 @@ pub(crate) trait Parser: Sized {
 #[macro_export]
 macro_rules! make_op {
     ($enum:ident, $(#[$outer:meta])* { $($operator:ident,)+ }) => {
-        #[derive(Clone, Debug, Display, PartialEq, Node, Visit)]
+        #[automatically_derived]
```
JohnnyMorganz, line 79: Don't think this change is correct - this derive should still be here?

full-moon/src/ast/parsers.rs
```diff
@@ -1563,11 +1563,20 @@ cfg_if::cfg_if! {
...
-            let (state, base_type) = if let Ok((state, identifier)) = {
-                ParseIdentifier
+            // Singleton type info: `"yes" | "no"` in `(string) -> "yes" | "no"`, `"$$typeof"` in `{ ["$$typeof"]: number }`,
+            let (state, base_type) = if let Ok((state, string_singleton)) = ParseStringLiteral.parse(state) {
+                (state, TypeInfo::Basic(string_singleton))
+            // Singleton type info: `true` in `type X = Error & { handled: true }`
+            } else if let Ok((state, true_singleton)) = ParseSymbol(Symbol::True).parse(state) {
+                (state, TypeInfo::Basic(true_singleton))
+            } else if let Ok((state, false_singleton)) = ParseSymbol(Symbol::False).parse(state) {
+                (state, TypeInfo::Basic(false_singleton))
```
JohnnyMorganz, line 1573: Should these be changed to `TypeInfo::SingletonString` and `TypeInfo::SingletonBool` (or equivalent).

I don't know if there is any downstream consumer that does something like type inference, but this separation may make it easier to work with.

The Luau parser also does this (https://github.com/Roblox/luau/blob/4b96f7efc12f04fef2f9511811c807bb30dedaef/Ast/include/Luau/Ast.h#L1184-L1196) so it may be worthwhile following what they do

full-moon/src/ast/parser_util.rs
matthargett: It was a suggestion on resolving the new clippy warnings that didn't pan out. Reverted.

full-moon/src/ast/parsers.rs
matthargett: I tried doing this locally, and it has ripple effects relating to StringLiteralQuoteType that appear to be far-reaching. I'm not against the idea, but would rather handle that in a separate PR.

@jeparlefrancais do you use full-moon for inference in darklua? Given the unique aspects of Luau's type system, especially in complex scenarios, means that reimplementing it in a way that doesn't trigger false positives or incorrect auto-completion seems slim.

JohnnyMorganz: I meant for it to still be defined as `TypeInfo::String(TokenReference)` or `TypeInfo::Boolean(TokenReference)` - the same as how Value is defined for a String/Number/Boolean in an expression: https://github.com/Kampfkarren/full-moon/blob/7f6ae19a7d2eb2d0ab941c8712518d1885d1a71c/full-moon/src/ast/mod.rs#L333-L338


Although maybe it's just minor for no benefit? Could just pattern match on the TokenReference for TypeInfo::Basic

jeparlefrancais: @matthargett No unfortunately darklua has its own AST for mutations and manipulations

jeparlefrancais: I think since TypeInfo wasn't created in this PR, we can probably leave it like that for now. 

It could be refactored in another PR though, I think it's good to have unique variants because you pick a single one if you only need one. If an enum has a variant that is a group (like `Basic` or `Symbol`) then the pattern matching becomes a lot uglier if you need to match on the group first, and then match again on TokenReference to know what kind of token it.

It's simpler if it's all separated, because if you need multiple you can just match multiple into a single pattern branch.

JohnnyMorganz (approved):
Can optionally look to extend TypeInfo for the singletons if we want (either in this PR or another one), but otherwise it all looks good

JohnnyMorganz (approved):

full-moon/src/ast/type_visitors.rs
```diff
@@ -126,6 +132,8 @@ impl VisitMut for TypeInfo {
                 TypeInfo::Array { braces, type_info }
             }
             TypeInfo::Basic(__self_0) => TypeInfo::Basic(__self_0.visit_mut(visitor)),
+            TypeInfo::Boolean(__self_0) => TypeInfo::Basic(__self_0.visit_mut(visitor)),
+            TypeInfo::String(__self_0) => TypeInfo::Basic(__self_0.visit_mut(visitor)),
```
JohnnyMorganz, line 136:
```suggestion
            TypeInfo::Boolean(__self_0) => TypeInfo::Boolean(__self_0.visit_mut(visitor)),
            TypeInfo::String(__self_0) => TypeInfo::String(__self_0.visit_mut(visitor)),
```

JohnnyMorganz (changes requested):
Very small bug, meant the snapshots were still saying `Basic` instead of string/boolean

Kampfkarren:
Will release and such later today

PR #222, opened 2022-02-16
Rewrite tokenizer

Rerumu:
Replace `peg` with `logos` for tokenizing. Avoids magic `peg` macros and provides ~2x more throughput on the lower end.


full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
...
+            let trim = &slice[count + 2..slice.len() - count - 2];
+
+            (trim.into(), Some(count))
+        }
+        None => (slice.into(), None),
+    }
+}
+
+fn test_bracket_head(slice: &str) -> Option<usize> {
+    // starts with `[`?
```
Kampfkarren, line 17: These comments are useless, I'd just ditch them.

full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
...
+    if !matches!(slice.chars().nth(count + 1), Some('[')) {
+        return None;
+    }
+
+    Some(count)
+}
+
+fn read_bracketed(lex: &mut Lexer<Atom>, skips: usize) -> bool {
+    let num_eq = match lex.slice().get(skips..).and_then(test_bracket_head) {
+        Some(v) => v,
```
Kampfkarren, line 32: Write out the value completely

full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
...
+    // ends with `[`?
+    if !matches!(slice.chars().nth(count + 1), Some('[')) {
+        return None;
+    }
+
+    Some(count)
+}
+
+fn read_bracketed(lex: &mut Lexer<Atom>, skips: usize) -> bool {
+    let num_eq = match lex.slice().get(skips..).and_then(test_bracket_head) {
```
Kampfkarren, line 31: Instead of using `and_then(test_bracket_head)`, just run `test_bracket_head(value)` inside the `Some` branch.

full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
...
+        Some(v) => v,
+        None => return false,
+    };
+
+    let mut search = false;
+    let mut num = 0;
+
+    for (i, v) in lex.remainder().char_indices() {
```
Kampfkarren, line 39: Full variable names.

full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
...
+    Some(count)
+}
+
+fn read_bracketed(lex: &mut Lexer<Atom>, skips: usize) -> bool {
+    let num_eq = match lex.slice().get(skips..).and_then(test_bracket_head) {
+        Some(v) => v,
+        None => return false,
+    };
+
+    let mut search = false;
```
Kampfkarren, line 36: Search what?

full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
...
+    Until,
+
+    #[token("while")]
+    While,
+
+    #[cfg(feature = "lua52")]
+    #[token("goto")]
+    Goto,
+
+    #[token("+=")]
```
Kampfkarren, line 127: These are all Luau specific, and should be behind the feature flag.

If they weren't before, I forget the reason for it, and I guess we'll see.

full-moon/src/atom.rs
```diff
@@ -0,0 +1,291 @@
+use logos::{Lexer, Logos};
+
+use crate::ShortString;
+
+pub(crate) fn trim_bracket_head(slice: &str) -> (ShortString, Option<usize>) {
```
Kampfkarren, line 5: Should just need to be `pub fn`

full-moon/src/tokenizer.rs
```diff
@@ -1,83 +1,322 @@
...
+    Star,
+
+    #[cfg_attr(feature = "serde", serde(rename = "~="))]
+    TildeEqual,
+}
+
+impl TryFrom<Atom> for Symbol {
+    type Error = ();
+
+    fn try_from(v: Atom) -> Result<Self, Self::Error> {
```
Kampfkarren, line 169: Full name

full-moon/src/tokenizer.rs
```diff
@@ -1,83 +1,322 @@
...
+        let ok = match v {
```
Kampfkarren, line 170: Change this entire thing to just `match v {` etc, and `Ok(...)` for everything.

Kampfkarren: Or `Ok(match v {` etc etc with `return Err(())`

Kampfkarren (changes requested):
No single letter variables please

full-moon/src/atom.rs
Rerumu: In this case `test_bracket_head` returns an `Option`, so running it with `and_then` ensures both the slice is valid and that it has a proper head.

full-moon/src/atom.rs
```diff
@@ -0,0 +1,295 @@
+use logos::{Lexer, Logos};
+
+use crate::ShortString;
+
+pub fn trim_bracket_head(slice: &str) -> (ShortString, Option<usize>) {
```
JohnnyMorganz, line 5: Visibility: is this necessary?
```suggestion
pub(crate) fn trim_bracket_head(slice: &str) -> (ShortString, Option<usize>) {
```

full-moon/src/tokenizer.rs
```diff
@@ -1,83 +1,327 @@
...
+    cmp::Ordering,
+    convert::{TryFrom, TryInto},
+    fmt::{self, Display},
+};
+
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "lowercase"))]
+#[allow(missing_docs)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Symbol {
```
JohnnyMorganz, line 20: Missing `non_exhaustive` attribute: https://github.com/Kampfkarren/full-moon/blob/main/full-moon-derive/src/symbols.rs#L98

JohnnyMorganz: Also missing the doc comment above it too: `/// A literal symbol, used for both words important to syntax (like while) and operators (like +)`

JohnnyMorganz (approved):

Kampfkarren:
There's still some stuff but it can get cleaned up

Kampfkarren:
@Rerumu 

1. Can you put this in the changelog?
2. Can you undo your changes to turn `formatter` into `f`?

PR #224, opened 2022-02-16
Support indexed types with type pack params

### LINKED ISSUE #223: variadic generic syntax not accepted on RHS of type assignment
```lua
--!strict
local Util = require(script.Parent.u)

export type Signal<T...> = Util.Signal<T...>
```

image

ZoteTheMighty:
Closes #223.

I can't tell what's going on at the top of the snapshot file; I'm not sure what I changed that would make it different. I don't fully understand the fix here; I mimicked another recent change from @JohnnyMorganz that seemed applicable. If this is off-base or has some consequence I'm not aware of, please let me know and I'm happy to dig a little deeper.


JohnnyMorganz (approved):
Changelog just needs updating

PR #225, opened 2022-03-06
Show feature gates on documentation

JohnnyMorganz:
Provides better documentation hints on the docs.rs page for nodes gated behind feature flags.
Example:
image


JohnnyMorganz:
Failure seems to be unrelated (clippy warning)

Kampfkarren:
Reminding me we need to move "roblox" -> "luau" 😅

PR #228, opened 2022-06-24
Fix string escape issue

memothelemo:
Both regexes [`ApostropheString`](https://github.com/Kampfkarren/full-moon/blob/main/full-moon/src/atom.rs#L276) and [`QuoteString`](https://github.com/Kampfkarren/full-moon/blob/main/full-moon/src/atom.rs#L279) doesn't end the string after the escaped sequence. Instead, it goes to further than the expected until it has a string quote with no backslash before it.

This is example of what I mean the result here this followed link:
https://regex101.com/r/D9BMpL/2

Instead of using regex as a solution for parsing strings, I made a function that consumes the escape string sequence and it ends
with the string quote regardless it has escaped sequence beside them.


JohnnyMorganz:
Can you add a test case?

Kampfkarren:
Update CHANGELOG as well please to describe the fix

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- Fixed issue with lexing strings which it matches the wrong escape sequence.
```
Kampfkarren, line 9:
```suggestion
- Fixed issue with lexing strings when it matches the wrong escape sequence.
```

PR #230, opened 2022-08-07
Fix clippy warnings

JohnnyMorganz:
Cleanup some derefs and a format_push_str.

Note: clippy wants `Eq` defined everywhere that `PartialEq` is - not 100% sure about this


Kampfkarren:
> Note: clippy wants Eq defined everywhere that PartialEq is - not 100% sure about this

Yeah, it's fine. I can't see us ever using floats in this codebase anyway.

PR #231, opened 2022-08-07
Fix panic on `TypeDeclaration::new()`

### LINKED ISSUE #229: `TypeDeclaration::new()` panics
Calling `TypeDeclaration::new()` will always panic, because it tries to call `TokenReference::symbol("type ")` which is not a symbol.  https://github.com/Kampfkarren/full-moon/blob/main/full-moon/src/ast/types.rs#L350

Whoops

JohnnyMorganz:
Fixes #229


Kampfkarren:
Tests fail :eyes:

JohnnyMorganz:
Was clippy which was failing, will rebase with master

PR #233, opened 2022-09-03
Add support for Lua 5.3

### LINKED ISSUE #221: [feature request] lua 5.3 support
What are the chances of adding support for lua 5.3? This seems fairly presumptuous on my part so fair enough if you close this.

JohnnyMorganz:
Lua 5.3 introduces new binary and unary operators (for binary expressions and integer division)
- `&`
- `|`
- `>>`
- `<<`
- `~` [also a unary operator]
- `//`

Closes #221


full-moon/src/tokenizer.rs
```diff
@@ -108,6 +108,10 @@ pub enum Symbol {
     #[cfg_attr(feature = "serde", serde(rename = ">"))]
     GreaterThan,
 
+    #[cfg(feature = "lua53")]
+    #[cfg_attr(feature = "serde", serde(rename = ">>"))]
+    RightShift,
```
JohnnyMorganz, line 113: I don't know if we should label `>>` as RightShift (and `<<` as LeftShift), as it seems most symbols aren't named after the function they provide (which makes sense, since if e.g. Luau introduces `>>` which isn't a RightShift operator)

The only other name I could think of is `DoubleGreaterThan`

JohnnyMorganz:
CI is red because of #234

full-moon/src/tokenizer.rs
Kampfkarren: DoubleGreaterThan is better IMO

Kampfkarren:
@JohnnyMorganz Merge from upstream please

PR #236, opened 2022-09-11
Add support for Lua 5.4

### LINKED ISSUE #232: [feature request] local variable attributes (Lua 5.4)
Lua 5.4 added support for `<const>` and `<close>` as attributes to local variables. I think just those two. https://www.lua.org/manual/5.4/manual.html#3.3.7

And it looks in usage like

```lua
local foo <const> = 3
```

Would you consider adding support for these? Even if the other 5.3 or 5.4 features aren't there, just this one would be useful for me.

JohnnyMorganz:
Closes #232

### DIFF
--- a/full-moon/src/ast/visitors.rs
+++ b/full-moon/src/ast/visitors.rs
@@ -446,11 +446,11 @@ impl VisitMut for LocalAssignment {
             let parameter = parameter_tuple.0.visit_mut(visitor);
             let attribute = attributes
                 .next()
-                .and_then(|attribute| attribute)
+                .flatten()
                 .map(|attribute| attribute.visit_mut(visitor));
             let type_specifier = type_specifiers
                 .next()
-                .and_then(|type_specifier| type_specifier)
+                .flatten()
                 .map(|type_specifier| type_specifier.visit_mut(visitor));
 
             let punctuation = parameter_tuple.1.visit_mut(visitor);
--- a/full-moon/src/ast/visitors.rs
+++ b/full-moon/src/ast/visitors.rs
@@ -370,7 +370,19 @@ impl Visit for LocalAssignment {
         visitor.visit_local_assignment(self);
         self.local_token.visit(visitor);
 
+        let mut attributes;
         let mut type_specifiers;
+
+        #[cfg(feature = "lua54")]
+        {
+            attributes = self.attributes();
+        }
+
+        #[cfg(not(feature = "lua54"))]
+        {
+            // TODO: Option<!>, and implement Visit for !
+            attributes = std::iter::repeat::<Option<Self>>(None);
+        }
 
         #[cfg(feature = "roblox")]
         {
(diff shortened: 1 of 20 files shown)


JohnnyMorganz:
The parsing of LocalAssignment is somewhat messy with the combination of lua 5.4 attributes and luau type specifiers, but it works. Let me know your thoughts

Kampfkarren:
It's ugly buuut I don't think we can get any better :(

full-moon/src/ast/visitors.rs
```diff
@@ -370,8 +370,20 @@ impl Visit for LocalAssignment {
         visitor.visit_local_assignment(self);
         self.local_token.visit(visitor);
 
+        let mut attributes;
```
Kampfkarren, line 373: You shouldn't need mut here (you don't need it on type_specifiers either).

full-moon/src/ast/visitors.rs
```diff
@@ -399,8 +412,20 @@ impl VisitMut for LocalAssignment {
         self = visitor.visit_local_assignment(self);
         self.local_token = self.local_token.visit_mut(visitor);
 
+        let mut attributes;
```
Kampfkarren, line 415: Same

full-moon/src/ast/visitors.rs
```diff
@@ -412,24 +437,35 @@ impl VisitMut for LocalAssignment {
...
+        let mut new_attributes = Vec::new();
         let mut new_type_specifiers = Vec::new();
         let mut new_names = Punctuated::new();
 
         for parameter_pair in self.name_list.into_pairs() {
             let parameter_tuple = parameter_pair.into_tuple();
             let parameter = parameter_tuple.0.visit_mut(visitor);
+            let attribute = attributes
+                .next()
+                .and_then(|attribute| attribute)
```
Kampfkarren, line 449: Why do we do this? Can we just `.flatten()` instead?

full-moon/src/ast/visitors.rs
JohnnyMorganz: Its necessary because `attributes` and `type_specifiers` are iterators which we mutate by calling `.next()` on later
https://github.com/Kampfkarren/full-moon/blob/aae3a2339bdbe3d677cd3f60775ed88aadb49f25/full-moon/src/ast/visitors.rs#L398-L402

full-moon/src/ast/visitors.rs
JohnnyMorganz: Not sure. Changed

full-moon/src/ast/visitors.rs
JohnnyMorganz: Ditto above, we mutate the iterator

will:
Awesome! Thanks @JohnnyMorganz and @Kampfkarren for merging

PR #238, opened 2022-09-15
Run all flags testing in CI

JohnnyMorganz:
Test all flags enabled at once to pick up potential conflicts


JohnnyMorganz:
We could potentially also get rid of the `roblox` ignore on this test:
https://github.com/Kampfkarren/full-moon/blob/19f98b778531e2d5d807d979bba6996040df2042/full-moon/tests/pass_cases.rs#L56-L60
Will need a bit more snapshot cleaning though. Let me know if I should do that

Kampfkarren:
Did you confirm that this would've caught the `>>` bug?

JohnnyMorganz:
Yeah it would have (I deleted the test and put it in the other PR)
The Lua 5.1 tests are ignored though, so they might not be picked up

PR #250, opened 2022-10-23
Use single quote for char type

saguywalker:


Kampfkarren:
Surprised I never saw this, sorry!

PR #252, opened 2022-10-30
Lua 5.2+: Support `\z` escape sequences in strings

### LINKED ISSUE #251: Lua 5.2: does not support \z string breaks
https://github.com/JohnnyMorganz/StyLua/issues/613

JohnnyMorganz:
`\z` escape sequences can be followed by whitespace (including whitespace) which is ignored
Fixes #251

### DIFF
--- a/full-moon/tests/lua52_cases/pass/z-escape-string/ast.snap
+++ b/full-moon/tests/lua52_cases/pass/z-escape-string/ast.snap
@@ -3,106 +3,76 @@ source: full-moon/tests/pass_cases.rs
 expression: ast.nodes()
 ---
 stmts:
-  - - LocalAssignment:
-        local_token:
-          leading_trivia: []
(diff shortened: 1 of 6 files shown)


full-moon/tests/lua52_cases/pass/z-escape-string/ast.snap
```diff
@@ -0,0 +1,108 @@
...
+                        bytes: 14
+                        line: 1
+                        character: 15
+                      end_position:
+                        bytes: 39
+                        line: 2
+                        character: 14
+                      token_type:
+                        type: StringLiteral
+                        literal: "testing \\z\n\t\t\t   twelve"
```
JohnnyMorganz, line 104: We should be including the whitespace even though it gets ignored at runtime, right?

Its important for the lossless-ness

PR #253, opened 2022-11-11
String interpolation

### LINKED ISSUE #235: Luau string interpolation
https://github.com/Roblox/luau/pull/614

Kampfkarren:
Adds support for string interpolation. Closes #235


matthargett:
What's this currently waiting on?

Kampfkarren:
Not much, I wanted @JohnnyMorganz 's review on it. If string interpolation comes out before then I'm satisfied enough with this.

JohnnyMorganz (commented):
Still plan on taking a deeper look at the actual implementation, but one thing that might be useful is more fail test case snapshots. e.g.:

```lua
local _ = `asdf{` -- currently errors with unclosed string. maybe the error message could be improved? but its reasonable enough
local _ = `asdf{}` -- this currently passes, but it should fail. Luau errors with "SyntaxError: Malformed interpolated string, expected expression inside '{}"
```

PR #258, opened 2022-12-07
Lua 5.2 / LuaJIT: Support fractional hex + binary exponents, and LuaJIT suffixes

JohnnyMorganz:
Fractional hex / hex with binary exponents are Lua 5.2 features

LuaJIT suffixes (LL / ULL / i) are also supported, but currently under the lua52 feature flag for simplicity.

If both lua52 and roblox features are enabled, we permit an amalgamation of them all so that it can all be parsed. This will allow false negatives, but it is the best that we can do for now


matthargett (approved):
If all of openresty lua parses correctly, this LGTM

PR #262, opened 2023-01-18
chore: fix clippy

attila-lin:
Have issue with 

```rust
pub enum Pair<T>
```

Do we need to limit `T` with `Display`


Kampfkarren:
Thanks!

PR #266, opened 2023-01-30
fix: lua5.2 and above support `\` for escaped too

attila-lin:


attila-lin:
I think it's clippy's bug.

The derive part is still `edition=2018`.

I can upgrade it if you are OK with it.

PR #270, opened 2023-02-25
Fix tokenizer positions when multiple newlines follow eachother

JohnnyMorganz:
There is a bug in `read_position` when multiple newlines follow each other, which manifests into the errors seen in https://github.com/evaera/moonwave/actions/runs/3990359990/jobs/6843912815

This is because the line number is not bumped for consecutive newline characters, it is only bumped once a non-newline character is seen.
We fix this by always bumping when `has_newline` is set, then checking afterwards for the `\n` character and setting `has_newline = true`


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed parse failed with Chinese token in comment and bump deps.
 - Added `first` method to `Punctuated`.
```
Kampfkarren, line 11: oops,we need to move this

Issue #265, opened 2023-01-19
misleading error given when Luau if-expression is missing `else` 


matthargett:
In this code:
```lua
exports.slice =  function<T>(array: Array<T>, startIndex_: number?, endIndex_: number?): Array<T>
        local length = #array

        -- JS translates indices > length to be length; in Lua, we translate to length + 1
        local endIndex = if (nil == endIndex_ or endIndex_ > length + 1)
                then length + 1
                else if endIndex_ < 1
                        then math.max(1, length - math.abs(endIndex_))
end
```

I get this message, which had me looking at closures above that I thought maybe I hadn't closed off yet:
```
error[parse_error]: unexpected token `if`
    ┌─ lib/array.lua:101:26
    │
101 │         local endIndex = if (nil == endIndex_ or endIndex_ > length + 1)
    │                          ^^ expected expression
```

It would be nicer if it precisely said `if expression is missing mandatory else clause` or something similar.

PR #276, opened 2023-05-05
Luau Parser rewrite

JohnnyMorganz:
Builds off #274 to add Luau support


full-moon/src/ast/parsers.rs
```diff
@@ -1121,6 +1255,44 @@ fn expect_while_stmt(
...
+#[cfg(feature = "roblox")]
+fn expect_type_declaration(
+    state: &mut ParserState,
+    type_token: TokenReference,
+) -> Result<ast::TypeDeclaration, ()> {
+    let base = match state.current()? {
+        token if token.token_kind() == TokenKind::Identifier => state.consume().unwrap(),
+        token => {
+            state.token_error(token.clone(), "expected type name");
+            // rewrite todo: maybe we can add an error name here to continue parsing?
```
Kampfkarren, line 1267: Probably not worth it, the end result from an errors perspective would be the same

full-moon/src/ast/parsers.rs
```diff
@@ -1663,24 +1926,935 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+    };
+
+    let ParserResult::Value(if_expression) = parse_expression(state) else {
+        return Err(())
+    };
+
+    let mut else_if_expressions = Vec::new();
+    while let Some(else_if_token) = state.consume_if(Symbol::ElseIf) {
+        let ParserResult::Value(condition) = parse_expression(state) else {
+            return Err(())
```
Kampfkarren, line 1986: Hmm, we could maybe recover these but then it'd just error when it eventually hits the `else` anyway

full-moon/src/ast/parsers.rs
```diff
@@ -1663,24 +1926,935 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+    let ParserResult::Value(simple_type) = parse_simple_type(state, false) else {
+        return ParserResult::LexerMoved;
+    };
+
+    parse_type_suffix(state, simple_type)
+}
+
+#[cfg(feature = "luau")]
+fn parse_type_or_pack(state: &mut ParserState) -> ParserResult<ast::TypeInfo> {
+    let ParserResult::Value(simple_type) = parse_simple_type(state, true) else {
```
Kampfkarren, line 2037: Same feedback here, marker enum instead of true/false

full-moon/src/ast/parsers.rs
```diff
@@ -1663,24 +1926,935 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+                    let Some(colon) = state.require(Symbol::Colon, "expected `:` after type field key") else {
+                        return Err(());
+                    };
+                    let value = match parse_type(state) {
+                        ParserResult::Value(value) => value,
+                        ParserResult::LexerMoved | ParserResult::NotFound => {
+                            state.token_error(
+                                state.current().unwrap().clone(),
+                                "expected type after type field key",
+                            );
```
Kampfkarren, line 2447: I think we don't usually throw an error if it is LexerMoved, right?

full-moon/src/ast/parsers.rs
```diff
@@ -1663,24 +1926,935 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+            }
+
+            Err(()) => {
+                fields.push(Pair::End(field));
+                break;
+            }
+        };
+    }
+
+    let Some(right_brace) = state.require(Symbol::RightBrace, "expected `}` to close type table") else {
```
Kampfkarren, line 2480: We can probably recover this as long as there is type info

full-moon/src/ast/parsers.rs
```diff
@@ -1663,24 +1926,935 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+        {
+            Some((state.consume().unwrap(), state.consume().unwrap()))
+        } else {
+            None
+        };
+
+        let type_info = match parse_type(state) {
+            ParserResult::Value(return_type) => return_type,
+            ParserResult::LexerMoved | ParserResult::NotFound => return Err(()),
+        };
```
Kampfkarren, line 2543: `let else` would be nice here

full-moon/src/tokenizer/lexer.rs
```diff
@@ -650,6 +738,26 @@ impl Lexer {
             }
 
             '-' => {
```
Kampfkarren, line 740: Maybe I'm missing it, but I don't see `..=`

JohnnyMorganz: It is present earlier in the file https://github.com/Kampfkarren/full-moon/pull/276/files#diff-48f263edf3a28e94ca0dcbed94e5b2c8b0e1d87ca59ad1527218eb005c73cf22R708-R720

full-moon/src/ast/parsers.rs
JohnnyMorganz: I think luau might be able to recover here, but I will leave it for now. If it seems useful later we can add it

full-moon/src/ast/parsers.rs
JohnnyMorganz: Yeah i'm not too sure about this. If there's an example of something that would be nice if recoverable then I can take a look

JohnnyMorganz: Actually, https://github.com/Kampfkarren/full-moon/issues/265 is a similar example (except at the else block) where this is present.

I think we need some way to add "error terminals" to the graph. Like AstExprError or AstTypeError

JohnnyMorganz:
All test cases passing! 🎉

full-moon/src/ast/parsers.rs
```diff
@@ -1121,6 +1292,44 @@ fn expect_while_stmt(
...
+    };
+
+    let equal_token = state
+        .require(Symbol::Equal, "expected `=` after type name")
+        .unwrap_or_else(|| TokenReference::basic_symbol("="));
+
+    let declare_as = match parse_type(state) {
+        ParserResult::Value(declare_as) => declare_as,
+        _ => return Err(()),
+    };
```
chriscerie, line 1322:
```suggestion
    let ParserResult::Value(declare_as) = parse_type(state) else {
        return Err(());
    };
```

full-moon/src/ast/parsers.rs
```diff
@@ -1452,7 +1663,68 @@ fn parse_primary_expression(state: &mut ParserState) -> ParserResult<Expression>
...
+
+    match expression {
+        #[cfg(feature = "luau")]
+        ParserResult::Value(expression) if state.lua_version().has_luau() => {
+            if let Some(assertion_op) = state.consume_if(Symbol::TwoColons) {
+                let cast_to = match parse_type(state) {
+                    ParserResult::Value(cast_to) => cast_to,
+                    _ => return ParserResult::LexerMoved,
+                };
```
chriscerie, line 1714:
```suggestion
                let ParserResult::Value(cast_to) = parse_type(state) else {
                    return ParserResult::LexerMoved;
                };
```

full-moon/src/ast/parsers.rs
```diff
@@ -1663,129 +1995,1245 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
-        name,
-        #[cfg(feature = "lua54")]
-        attribute: None,
-    }
-}
+#[cfg(feature = "luau")]
+fn expect_interpolated_string(
+    state: &mut ParserState,
+    mut current: TokenReference,
+) -> Result<ast::InterpolatedString, ()> {
```
chriscerie, line 2112: Does this need to return Result? It looks like all codepaths return `Ok`.

full-moon/src/ast/parsers.rs
```diff
@@ -1663,129 +1995,1245 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+    let Ok(current_token) = state.current() else {
+        return ParserResult::NotFound;
+    };
+
+    if current_token.is_symbol(Symbol::Ellipse) {
+        let ellipse = state.consume().unwrap();
+        let type_info = match parse_type(state) {
+            ParserResult::Value(type_info) => type_info,
+            _ => return ParserResult::LexerMoved,
         };
```
chriscerie, line 2239:
```suggestion
        let ParserResult::Value(type_info) = parse_type(state) else {
            return ParserResult::LexerMoved;
        };
```

full-moon/src/ast/parsers.rs
```diff
@@ -1663,129 +1995,1245 @@ fn parse_function_body(state: &mut ParserState) -> ParserResult<FunctionBody> {
...
+                error_token()
+            }
+
+            Err(()) => return ParserResult::LexerMoved,
+        };
+
+        let current_token = match state.current() {
+            Ok(token) => token,
+            Err(()) => return ParserResult::NotFound,
+        };
```
chriscerie, line 2884:
```suggestion
        let Ok(current_token) = state.current() else {
            return ParserResult::NotFound;
        };
```

chriscerie (commented):
Looks great overall

full-moon/src/ast/parsers.rs
JohnnyMorganz: Good spot, looks unnecessary

PR #288, opened 2023-10-07
bump `indexmap`  to 2.0

attila-lin:
fix warning issue of `resolver`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,3 +4,5 @@ members = [
 	"full-moon",
 	"full-moon-derive",
 ]
+
+resolver = "2"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -3,4 +3,6 @@
 members = [
 	"full-moon",
 	"full-moon-derive",
-]
\ No newline at end of file
+]
+
+resolver = "2"
```
Kampfkarren, line 8: We shouldn't need this, right? `edition = "2021"` implies resolver = "2"

Kampfkarren: Oh, lame...

```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

PR #293, opened 2023-11-23
Support floor division assignment (//=)

### LINKED ISSUE #292: Missing floor division compound operator `//=`
I saw that the floor division operator was added recently, but the compound assign statement is missing the equivalent operator:

```lua
value //= 2
```

chriscerie:
Closes #292.


full-moon/src/ast/parsers.rs
```diff
@@ -2260,6 +2260,7 @@ make_op_parser!(CompoundOp, ParseCompoundOp,
         MinusEqual,
         StarEqual,
         SlashEqual,
+        DoubleSlashEqual,
```
Kampfkarren, line 2263: Standard is `TwoSlashEqual`

Kampfkarren (changes requested):

full-moon/src/ast/parsers.rs
chriscerie: It’d be ideal if it’s consistent with its counterpart `DoubleSlash`. Should that be changed to `TwoSlash` then?

jeparlefrancais (approved):
Thank you for jumping on this 🥇

full-moon/src/ast/parsers.rs
JohnnyMorganz: Ah i made a mistake when naming double slash originally. That's annoying...

Yeah I think it's better to stay related to DoubleSlash. Renaming to TwoSlash is a breaking change

Kampfkarren: Fair enough then don't change anything but we should deprecate both at some point

PR #297, opened 2024-04-07
Fix incorrect unary precedence

### LINKED ISSUE #296: Incorrect unary precedence
The rewritten parser incorrectly treats unary precedence, thus parsing the following:
`local c = -a + b`

incorrectly gives:
`UnaryOperator(BinaryOperator(a, b))`
instead of the expected
`BinaryOperator(UnaryOperator(a), b)`

chriscerie:
Fixes #296.

### DIFF
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -2074,6 +2074,11 @@ macro_rules! make_bin_op {
                     }
                 }
 
+                /// The precedence of unary operators.
+                pub fn precedence_of_unary() -> u8 {
+                    11
+                }
+
                 /// The token associated with this operator
                 pub fn token(&self) -> &TokenReference {
                     match self {
--- a/full-moon/src/ast/mod.rs
+++ b/full-moon/src/ast/mod.rs
@@ -2055,7 +2055,8 @@ macro_rules! make_bin_op {
             }
 
             impl BinOp {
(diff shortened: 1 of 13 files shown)


full-moon/src/ast/mod.rs
```diff
@@ -2074,6 +2074,11 @@ macro_rules! make_bin_op {
                     }
                 }
 
+                /// The precedence of unary operators
+                pub fn precedence_of_unary() -> u8 {
+                    11
+                }
```
chriscerie, line 2080: This seems a bit weird. Is there a better approach?

full-moon/src/ast/parsers.rs
```diff
@@ -1815,7 +1815,26 @@ fn parse_unary_expression(
...
+            );
+            return ParserResult::NotFound;
+        }
+        ParserResult::LexerMoved => return ParserResult::LexerMoved,
+    };
+
+    let expression = match parse_expression_with_precedence(
+        state,
+        primary_expression,
+        ast::BinOp::precedence_of_unary(),
```
JohnnyMorganz, line 1836: nit: why not just add `precedence()` to UnOp instead?

JohnnyMorganz (approved):
LGTM (I did not look at the snapshots in detail as GitHub doesn't display it in a useful way, will try to look in better detail in code editor later)

Can you also add the test cases that I initially sent on Discord?

PR #300, opened 2024-06-09
Support LuaJIT

JohnnyMorganz:
LuaJIT parsing was not implemented in the parser rewrite.

We add in support, now under a separate feature flag `luajit`. `LuaVersion::luajit()` can enable this.


full-moon/src/tokenizer/lexer.rs
```diff
@@ -1033,7 +1041,7 @@ impl Lexer {
         binary_character: char,
         start_position: Position,
     ) -> Option<LexerResult<Token>> {
-        debug_assert!(self.lua_version.has_luau());
+        debug_assert!(self.lua_version.has_luau() | self.lua_version.has_luajit());
```
Kampfkarren, line 1044: `||`?

Kampfkarren (approved):

PR #302, opened 2024-06-22
Ensure an error is emitted when parsing a stmt after a last stmt in the top level block

### LINKED ISSUE #298: laststmt behavior
test case:
```
print("xd")
continue -- this can be replaced with break/return
print("xd2")
```

parsing this does not produce any errors (imo it should).
however, even if this should be parsed, the location of the laststmt in question is ambiguous:
if it's continue, then we keep parsing statements, so the continue statement "teleports" to the end
if it's return, then the other statements get stripped (parser is no longer lossless)
etc. etc.

of course it's possible to determine the position based on tokens, but it isn't very convenient

JohnnyMorganz:
There is code that "merges" two top level blocks when performing initial parsing. This code does not account for the fact that a stmt parsed after a last stmt is invalid syntax. We extend it to ensure an error is reported in this case.

Note that we do not resolve the problem where the stmts after the last stmt "teleport" in front of the last stmt in the AST. That is more difficult to sort without restructuring the AST. However this was originally parsed as invalid syntax so we now at least emit an error and do not guarantee the state of an invalid syntax tree.

Partially resolves #298

### DIFF
--- a/full-moon/src/ast/parser_structs.rs
+++ b/full-moon/src/ast/parser_structs.rs
@@ -277,6 +277,8 @@ impl AstResult {
             _ => Block::new(),
         };
 
+        let block_has_last_stmt = block.last_stmt().is_some();
+
         loop {
             match parser_state.lexer.current() {
                 Some(LexerResult::Ok(token)) if token.token_kind() == TokenKind::Eof => {
@@ -313,6 +315,14 @@ impl AstResult {
                             }
 
                             continue;
+                        }
+
+                        if block_has_last_stmt {
+                            use crate::node::Node;
+                            parser_state.token_error(
+                                new_block.tokens().next().unwrap().clone(),
+                                "unexpected statement after last statement",
+                            )
                         }
 
                         block.merge_blocks(new_block);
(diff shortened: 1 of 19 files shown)


full-moon/src/ast/parser_structs.rs
```diff
@@ -315,6 +317,14 @@ impl AstResult {
                             continue;
                         }
 
+                        if block_has_last_stmt {
+                            use crate::node::Node;
```
Kampfkarren, line 321: Let's just put this in the beginning

Kampfkarren (approved):

full-moon/src/ast/parser_structs.rs
JohnnyMorganz: I was lazy because it clashes with start_position / end_position on TokenReference. If we stick it at the top we need to disambiguate all other usages

JohnnyMorganz: Done in last commit, wasn't that bad. Can revert if we want to

PR #313, opened 2024-08-04
implement table type access modifiers

jackdotink:
Implements table type access modifiers:
```luau
type foo = {
    read bar: number
    write baz: string,
}
```

### DIFF
--- a/full-moon/src/ast/luau.rs
+++ b/full-moon/src/ast/luau.rs
@@ -15,7 +15,13 @@ use derive_more::Display;
 #[non_exhaustive]
 pub enum TypeInfo {
     /// A shorthand type annotating the structure of an array: { number }
-    #[display(fmt = "{}{}{}", "braces.tokens().0", "type_info", "braces.tokens().1")]
+    #[display(
+        fmt = "{}{}{}{}",
+        "braces.tokens().0",
+        "display_option(access)",
+        "type_info",
+        "braces.tokens().1"
+    )]
     Array {
         /// The braces (`{}`) containing the type info.
         braces: ContainedSpan,
--- a/full-moon/src/ast/luau.rs
+++ b/full-moon/src/ast/luau.rs
@@ -26,6 +26,10 @@ pub enum TypeInfo {
         /// The braces (`{}`) containing the type info.
         braces: ContainedSpan,
         /// The access modifer of the array, `read` in `{ read number }`.
+        #[cfg_attr(
+            feature = "serde",
+            serde(default, skip_serializing_if = "Option::is_none")
+        )]
         access: Option<TokenReference>,
         /// The type info for the values in the Array
         type_info: Box<TypeInfo>,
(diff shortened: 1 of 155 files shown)


full-moon/src/ast/luau.rs
```diff
@@ -19,6 +19,8 @@ pub enum TypeInfo {
     Array {
         /// The braces (`{}`) containing the type info.
         braces: ContainedSpan,
+        /// The access modifer of the array, `read` in `{ read number }`.
+        access: Option<TokenReference>,
```
Kampfkarren, line 23: You don't have this in the `display`

Kampfkarren, line 23: Add `#[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]`

Kampfkarren: You didn't do this

Kampfkarren:
Need tests for this as well

full-moon/src/ast/luau.rs
jackdotink: I added `serde_option` or whatever it is to the entire enum. If you'd rather have this, I can do that instead.

DataM0del (commented):
This should require the [syntax property access modifiers](https://github.com/luau-lang/rfcs/blob/master/docs/syntax-property-access-modifiers.md) RFC to be implemented into the luau-lang/luau repository.
I don't think it's implemented yet, and I tried the example provided in the PR description and ran it with Luau, but it gave me a syntax error.

jackdotink:
@DataM0del try again because it works for me

DataM0del:
> @DataM0del try again because it works for me

I'll update my Luau version. I wish though that Luau would have a --version command but the PR for that isn't moving because of complaints about how you have to change the version macro value to the current version or etc.

DataM0del:
image
I updated. It works now.

DataM0del:
> @DataM0del try again because it works for me

Actually, it doesn't work because of the trailing comma.

full-moon/src/ast/luau.rs
Kampfkarren: Do you mean `display_option`? That's for display: the thing I gave is to not include it in serde if it's unset.

full-moon/src/ast/luau.rs
```diff
@@ -266,8 +274,9 @@ pub enum IndexedTypeInfo {
 /// The `foo: number` in `{ foo: number }`.
 #[derive(Clone, Debug, Display, PartialEq, Node, Visit)]
 #[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
-#[display("{key}{colon}{value}")]
+#[display("{}{key}{colon}{value}", display_option(access))]
 pub struct TypeField {
+    pub(crate) access: Option<TokenReference>,
```
Kampfkarren, line 279: This also needs the serde thing

PR #316, opened 2024-08-18
Fix clippy lints

jayvdb:


DataM0del (approved):
Why not? It's been here for months and no attention, although it just fixes clippy lints so that makes sense.

JohnnyMorganz (approved):
Thank you!

PR #317, opened 2024-10-01
Fix including trivia in check for typeof identifier

### LINKED ISSUE #314: Unable to parse intersection with `typeof` (rc5)
I found this issue with full-moon rc5, where intersection types using typeof don't parse:

```lua
export type Expect = typeof({}) & string
```

Which comes up in jest for example: https://github.com/jsdotlua/jest-lua/blob/main/src/expect/src/types.lua#L406-L408

YetAnotherClown:
Fixes #314

The following case will fail to parse correctly because the check at [this line](https://github.com/Kampfkarren/full-moon/blob/674154d21df9a367fed8b179167d9bd2ce811780/full-moon/src/ast/parsers.rs#L2348C13-L2348C46) includes all trivia. When including the trivia, we end up with a check that looks like ``"    typeof" = "typeof"``, which of course fails.
```lua
type T =
    typeof({})
```

This PR turns the token into a string instead of the token reference in the check, as the token does not contain trivia.

### DIFF
--- a/full-moon/tests/roblox_cases/pass/multiline_typeof_regression/ast.snap
+++ b/full-moon/tests/roblox_cases/pass/multiline_typeof_regression/ast.snap
@@ -1,0 +1,190 @@
+---
+source: full-moon/tests/pass_cases.rs
+expression: ast.nodes()
+input_file: full-moon/tests/roblox_cases/pass/multiline_typeof_regression
+---
+stmts:
+  - - TypeDeclaration:
+        type_token:
+          leading_trivia: []
+          token:
+            start_position:
+              bytes: 0
+              line: 1
+              character: 1
+            end_position:
+              bytes: 4
+              line: 1
+              character: 5
+            token_type:
+              type: Identifier
+              identifier: type
+          trailing_trivia:
+            - start_position:
+                bytes: 4
+                line: 1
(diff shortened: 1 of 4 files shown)


JohnnyMorganz (approved):
Can you add a test case for this?

full-moon/tests/roblox_cases/pass/multiline_typeof_regression/ast.snap
```diff
@@ -0,0 +1,190 @@
...
+                bytes: 13
+                line: 1
+                character: 14
+              end_position:
+                bytes: 15
+                line: 1
+                character: 15
+              token_type:
+                type: Whitespace
+                characters: "\r\n"
```
JohnnyMorganz, line 86: not sure if CI will be happy with these line endings but we'll see

JohnnyMorganz (approved):

YetAnotherClown:
@JohnnyMorganz The tests should be all good now, I had forgot to update `ast.snap` as well after fixing the `tokens.snap`, sorry about that!

DataM0del (approved):
It's just a one line change, shouldn't be bad.

JohnnyMorganz (approved):
Thank you!

PR #320, opened 2024-10-05
Add playdate support

DataM0del:
> [!NOTE]
> I'm starting this PR early because I haven't really interacted with this other than using its user-facing exported APIs and updating crates.


DataM0del:
oop...

DataM0del:
rebased stupidly... ignore this

DataM0del:
it's the email. Will rebase again.
context: I rebased so my email would be the one provided by GitHub. The users.github.com address I had in my gh.gitconfig wasn't correct, and that's why it didn't properly show as mine.

PR #321, opened 2024-10-05
feat: add playdate lua support

### LINKED ISSUE #282: Playdate runtime support
Hello. For the [Playdate](https://play.date/) console, a Lua 5.4 runtime is provided with some minor syntax changes:

https://sdk.play.date/inside-playdate/#playdate-sdk-lua-enhancements

I think most or all of these are already in the code but feature gated behind `roblox`. It would be nice to have a `playdate` configuration.

DataM0del:
Don't ask what happened on the previous PR (I set the wrong git email for my github gitconfig, because I use Codeberg btw, but I just leave my github config there just for contributions)...
Maybe one day I'll make a thing to manage all of this!
Fixes #282


Kampfkarren:
Are `+=` in Lua 5.4 and such? Looks like Playdate has them, but I don't know much about PUC-Rio Lua past 5.1.

If not, you'll want to expand the existing `+=` implementations (added for Roblox's Luau) to check Playdate.

matthargett:
Should this be taken out of draft?

DataM0del:
> Should this be taken out of draft?

No, as I haven't really done anything to make it actually work with some syntax and what not, mainly because I'm lazy and I don't know how to do stuff in this codebase.

PR #333, opened 2024-12-07
Fix panic during malformed empty interpolated string

### LINKED ISSUE #332: Empty interpolated string causes a panic
Attempting to parse an empty interpolating string (luau) causes an unrecoverable panic. Instead of crashing, it should return a failure result as empty interpolated strings are not valid luau syntax.

The associated line that causes the panic is [here](https://github.com/Kampfkarren/full-moon/blob/main/full-moon/src/ast/parsers.rs#L1701)

```luau
print(`{}`) 
--      ^
--      | crashes as soon as } is typed
```
```
thread 'notify-rs debouncer loop' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-6f17d22bba15001f\full_moon-1.1.0\src\ast\parsers.rs:1701:26:
internal error: entered unreachable code: unexpected interpolated string kind: End
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

JohnnyMorganz:
Fixes #332


full-moon/tests/roblox_cases/fail/parser/string_interpolation_empty/ast_to_string.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: full-moon/tests/fail_cases.rs
+expression: ast.to_string()
+input_file: full-moon/tests/roblox_cases/fail/parser/string_interpolation_empty
+---
+"print(`{)"
```
JohnnyMorganz, line 6: Not sure if I like this resolved AST to string

JohnnyMorganz:
Going to merge this, we can look into improving the fallible parsing later if it is useful for someone

PR #337, opened 2024-12-30
Support parsing Luau function attributes

### LINKED ISSUE #336: Support parsing function attributes
image
https://docs.rs/full_moon/1.1.2/src/full_moon/ast/parsers.rs.html#2571

The typed code includes unions and intersections, and does not have any code which directly uses the new solver.
I am also not sure if this is darklua's fault or full_moon's fault.

JohnnyMorganz:
Supports the parsing of Luau function attributes, such as `@native` or `@checked`. Does not perform any validation on the name of attributes, unlike Luau.

Unfortunately, due to the existing `Attribute` struct for Lua 5.4 `<const>` attributes, we have to name the struct `LuauAttribute`.

**[BREAKING CHANGE]**: Because attributes can be attached to anonymous functions, we create a new `AnonymousFunction` struct to handle this. This replaces the old `Box<(TokenReference, FunctionBody)>`.

Closes #336


jok-dev:
Would be great if this could be merged.

PR #338, opened 2025-02-09
Support CfxLua

mdxwzl:
- Implemented CFX Lua as a feature flag - `cfxlua`
  - Compound Operators: `+=, -=, *=, /=, <<=, >>=, &=, |=, and ^=`
  - Safe Navigation e.g. `t?.x?.y == nil`
  - In Unpacking e.g. `local a, b, c = t` is equivalent to `local a, b, c = t.a, t.b, t.c`
  - Set Constructors e.g. `t = { .x, .y }` is equivalent to `t = { x = true, y = true }`
  - C-Style Comments (single & multiline) e.g. `/* comment */`
  - Compile Time Jenkins' Hashes e.g. ``` `Hello, World!` -> 1395890823```


mdxwzl:
Some of the tests dont pass in the workflow, I'll look into it tmrw

mdxwzl:
Would be glad if someone can look over it. I tested the workflows locally with Docker, and it keeps changing up the bytes, line & character values in the AST. Somehow, they are not the same in the workflow environment as on my Windows machine.

mdxwzl:
Test should run now

full-moon/src/ast/luau.rs
```diff
@@ -893,114 +893,9 @@ impl ExportedTypeFunction {
...
-    }
-
-    /// Returns a new CompoundAssignment with the given value being assigned
-    pub fn with_rhs(self, rhs: Expression) -> Self {
-        Self { rhs, ..self }
-    }
-}
+/*
+    CompoundOp and CompoundAssignment have been moved to `compound.rs´, since CfxLua makes use of them as well.
+*/
```
JohnnyMorganz, line 898: Can we re-export the types here? To reduce the breaking change on downstream users

full-moon/src/ast/compound.rs
```diff
@@ -0,0 +1,140 @@
...
+#[display("{_0}")]
+/// Compound operators, such as X += Y or X -= Y
+pub enum CompoundOp {
+    PlusEqual(TokenReference),
+    MinusEqual(TokenReference),
+    StarEqual(TokenReference),
+    SlashEqual(TokenReference),
+    CaretEqual(TokenReference),
+
+    // luau sepcific
```
JohnnyMorganz, line 20: Should these be gated via `cfg(feature = "luau")` flags if they are only available in luau?

Some with the below for CfxLua

full-moon/src/ast/compound.rs
```diff
@@ -0,0 +1,140 @@
...
+    DoubleSlashEqual(TokenReference),
+    PercentEqual(TokenReference),
+    TwoDotsEqual(TokenReference),
+
+    // cfxlua sepcific
+    LeftShift(TokenReference),
+    RightShift(TokenReference),
+    BitwiseAndAssignment(TokenReference),
+    BitwiseOrAssignment(TokenReference),
```
JohnnyMorganz, line 29: To align with the other options. We try to make the names represent the characters rather than the logic, since in future `>>=` could mean something other than right shift operator for a different Lua dialect

```suggestion
    DoubleLessThanEqual(TokenReference),
    DoubleGreaterThanEqual(TokenReference),
    AmpersandEqual(TokenReference),
    PipeEqual(TokenReference),
```

full-moon/src/ast/mod.rs
```diff
@@ -39,7 +46,7 @@ mod versions;
 
 #[cfg(any(feature = "lua52", feature = "luajit"))]
 pub mod lua52;
-#[cfg(feature = "lua54")]
+#[cfg(any(feature = "lua54", feature = "cfxlua"))]
```
JohnnyMorganz, line 49: Since `cfxlua` already extends lua54, we don't need to change this config since lua54 should always be enabled

```suggestion
#[cfg(feature = "lua54")]
```

full-moon/src/ast/mod.rs
```diff
@@ -413,11 +430,11 @@ pub enum Stmt {
 
     /// A goto statement, such as `goto label`
     /// Only available when the "lua52" or "luajit" feature flag is enabled.
-    #[cfg(any(feature = "lua52", feature = "luajit"))]
+    #[cfg(any(feature = "lua52", feature = "luajit", feature = "cfxlua"))]
```
JohnnyMorganz, line 433: Same as above and in the other cases, I think this can be left alone
```suggestion
    #[cfg(any(feature = "lua52", feature = "luajit"))]
```

full-moon/src/ast/versions.rs
```diff
@@ -121,6 +122,25 @@ impl LuaVersion {
     pub fn has_luajit(self) -> bool {
         cfg!(feature = "luajit") && (self.bitfield & VERSION_LUAJIT != 0)
     }
+
+    /// Creates a new LuaVersion with only CFX Lua.
+    #[cfg(feature = "cfxlua")]
+    pub fn cfxlua() -> Self {
+        Self {
+            bitfield: VERSION_CFXLUA,
```
JohnnyMorganz, line 130: Similar to the `lua54` example above, since cfxlua is an extension of lua54, we should extend the bitfield
```suggestion
            bitfield: VERSION_LUA52 | VERSION_LUA53 | VERSION_LUA54 | VERSION_CFXLUA,
```

full-moon/src/tokenizer/structs.rs
```diff
@@ -121,21 +120,28 @@ symbol! {
...
 
         [luau | lua53] Ampersand => "&",
         [luau] ThinArrow => "->",
         [luau | lua52 | luajit] TwoColons => "::",
 
+        [cfxlua] LeftShift => "<<=",
+        [cfxlua] RightShift => ">>=",
+        [cfxlua] BitwiseAndAssignment => "&=",
+        [cfxlua] BitwiseOrAssignment => "|=",
+        [cfxlua] SafeNavigation => "?.",
```
JohnnyMorganz, line 143: As mentioned above

```suggestion
        [cfxlua] DoubleLessThanEqual => "<<=",
        [cfxlua] DoubleGreaterThanEqual => ">>=",
        [cfxlua] AmpersandEqual => "&=",
        [cfxlua] PipeEqual => "|=",
        [cfxlua] QuestionMarkDot => "?.",
```

To align with the other options. We try to make the names represent the characters rather than the logic, since in future >>= could mean something other than right shift operator for a different Lua dialect

full-moon/src/tokenizer/structs.rs
```diff
@@ -121,21 +120,28 @@ symbol! {
         Until => "until",
         While => "while",
 
-        [lua52 | luajit] Goto => "goto",
+        [lua52 | luajit | cfxlua] Goto => "goto",
```
JohnnyMorganz, line 123: If you make the above change to the bitfield, you won't need to change this
```suggestion
        [lua52 | luajit] Goto => "goto",
```

full-moon/src/tokenizer/structs.rs
```diff
@@ -390,6 +417,10 @@ pub enum TokenKind {
     #[cfg(feature = "luau")]
     /// Some form of interpolated string
     InterpolatedString,
+
+    #[cfg(feature = "cfxlua")]
+    /// A single/multi line C-style comment, such as `/* comment */`
+    CStylComment,
```
JohnnyMorganz, line 423: Typo, also we don't need to mention single/multi line, since it can always be multiline
```suggestion
    /// A C-style comment, such as `/* comment */`
    CStyleComment,
```

full-moon/src/visitors.rs
```diff
@@ -302,20 +302,26 @@ create_visitor!(ast: {
     }
 
     // Lua 5.2
-    #[cfg(any(feature = "lua52", feature = "luajit"))] {
+    #[cfg(any(feature = "lua52", feature = "luajit", feature = "cfxlua"))] {
```
JohnnyMorganz, line 305:
```suggestion
    #[cfg(any(feature = "lua52", feature = "luajit"))] {
```

full-moon/src/visitors.rs
```diff
@@ -302,20 +302,26 @@ create_visitor!(ast: {
...
         visit_goto => Goto,
         visit_label => Label,
     }
 
     #[cfg(feature = "lua54")] {
         visit_attribute => Attribute,
     }
+
+    #[cfg(all(feature = "cfxlua", not(feature = "luau")))] {
```
JohnnyMorganz, line 314: I don't quite understand this condition.

We can remove the `visit_compound_assignment` from the Luau section above and then use `any`

```suggestion
    #[cfg(any(feature = "cfxlua", feature = "luau"))] {
```

full-moon/src/visitors.rs
```diff
@@ -302,20 +302,26 @@ create_visitor!(ast: {
...
+        visit_compound_assignment => CompoundAssignment,
+        visit_compound_op => CompoundOp,
+    }
 }, token: {
     visit_identifier,
     visit_multi_line_comment,
     visit_number,
     visit_single_line_comment,
     visit_string_literal,
+    visit_c_style_comment,
```
JohnnyMorganz, line 324: Should be gated by a `cfxlua` check

full-moon/src/visitors.rs
```diff
@@ -6,7 +6,7 @@ use crate::{
 
 #[cfg(any(feature = "lua52", feature = "luajit"))]
 use crate::ast::lua52::*;
-#[cfg(feature = "lua54")]
+#[cfg(any(feature = "lua54", feature = "cfxlua"))]
```
JohnnyMorganz, line 9:
```suggestion
#[cfg(feature = "lua54")]
```

full-moon/src/ast/mod.rs
```diff
@@ -206,6 +213,16 @@ pub enum Field {
         value: Expression,
     },
 
+    /// A set constructor field, such as .a inside { .a } which is equivalent to { a = true }
+    #[display("{dot}{name}")]
+    #[cfg(feature = "cfxlua")]
+    SetConstructorField {
```
JohnnyMorganz, line 219: We are already in Field, we don't need `Field` at the end of the name
```suggestion
    SetConstructor {
```

full-moon/src/ast/parsers.rs
```diff
@@ -8,6 +8,10 @@ use super::{
     span::ContainedSpan,
     Expression, FunctionBody, Parameter,
 };
+
+#[cfg(any(feature = "cfxlua", feature = "luau"))]
+use super::Var;
```
JohnnyMorganz, line 13: Looks like this was introduced by your IDE.

You should reference it through `ast::Var` instead, which is already exported

JohnnyMorganz (commented):
Thanks for this! I had an initial look - I haven't looked deep into the parser / tokenizer changes yet, I will do this in a follow up

Kuuzoo:
Hi @JohnnyMorganz,
Thank you for your feedback!

I’ve just implemented the suggested changes last night and verified that Clippy is now passing too.

Let me know if there’s anything else you’d like to see.
Thanks again for reviewing!

full-moon/src/ast/compound.rs
```diff
@@ -0,0 +1,154 @@
...
+#[non_exhaustive]
+#[allow(missing_docs)]
+#[display("{_0}")]
+/// Compound operators, such as X += Y or X -= Y
+pub enum CompoundOp {
+    PlusEqual(TokenReference),
+    MinusEqual(TokenReference),
+    StarEqual(TokenReference),
+    SlashEqual(TokenReference),
+    CaretEqual(TokenReference),
```
Kampfkarren, line 18: IMO for this and the checks you should do `#[cfg(any(feature = "luau", feature = "cfxlua"))]`--it makes it clear when adding more what's actually supported by Lua itself (nothing) and what isn't.

full-moon/src/ast/luau.rs
```diff
@@ -893,114 +893,25 @@ impl ExportedTypeFunction {
...
-    pub fn with_rhs(self, rhs: Expression) -> Self {
-        Self { rhs, ..self }
-    }
-}
+/// A compound assignment operator, such as `+=`, `-=`, etc.
+/// This has been moved to `compound.rs` since CfxLua makes use of it as well.
+#[cfg(not(feature = "luau"))]
+#[cfg_attr(docsrs, doc(cfg(not(feature = "luau"))))]
+#[deprecated(
+    since = "full-moon 0.18.0",
```
Kampfkarren, line 901: We're way past this already, IMO I would ditch since entirely since it doesn't add much value

full-moon/src/tokenizer/lexer.rs
```diff
@@ -602,7 +602,30 @@ impl Lexer {
...
+                                        } else {
+                                            Vec::new()
+                                        },
+                                    )
+                                }
+                            }
+                        });
                     }
+
+
```
Kampfkarren, line 628: Nit: Remove these

full-moon/src/util.rs
```diff
@@ -11,6 +11,14 @@ pub fn empty_optional_vector<T>(vec: &Vec<Option<T>>) -> bool {
     vec.iter().all(Option::is_none)
 }
 
+// This allows to check if the vector is full of None's -> true
+// If the vector length is 0 or there is at least one Some -> false
+#[cfg(feature = "luau")]
+#[allow(clippy::ptr_arg)]
+pub fn no_luau_usage<T>(vec: &Vec<Option<T>>) -> bool {
```
Kampfkarren, line 18: I don't think this is a good name--`vec_empty_or_all_none`?

Kampfkarren (approved):

Issue #346, opened 2025-11-02
Unexpected stack overflow while assigning function


saaawdust:
The following code causes a stack overflow while attempting to parse on version `2.0.0`. I only have the `luau` feature enabled:
```lua
function Foo(bar: foobar.baz): foobar.qux
    t.Free = function()
        -- any content in here results in a stack overflow
        -- if the following line is commented, no stack overflow
        bar.free()
    end
end
```

Full code:
```rs
use full_moon;

fn main() {
    full_moon::parse("function Foo(bar: foobar.baz): foobar.qux
    t.Free = function()
        -- any content in here results in a stack overflow
        -- if the following line is commented, no stack overflow
        bar.free()
    end
end");
}
```

PR #355, opened 2026-05-06
Shrink AST nodes to fix recursive-parse stack overflows (#346)

wez:
The AST stored heavy nodes by value inside enums and structs.  Because an enum's size is its largest variant, this caused size inflation that propagated through the tree: Stmt was 4784 bytes per slot (dominated by NumericFor's three inline 880-byte Expressions), and the recursive parser carried several of these per stack frame. A trivial program could overflow the default thread stack.

Reproduces with this snippet under any thread that has less than ~2 MiB of stack:

    use full_moon;

    fn main() {
        std::thread::Builder::new()
            .stack_size(1024 * 1024)
            .spawn(|| {
                full_moon::parse(
                    "function Foo(bar: foobar.baz): foobar.qux
                         t.Free = function()
                             bar.free()
                         end
                     end",
                )
                .unwrap();
            })
            .unwrap()
            .join()
            .unwrap();
    }

For context: Rust's default `std::thread::spawn` stack is 2 MiB on tier-1 platforms; the Linux main thread gets 8 MiB; Tokio's multi-thread runtime workers default to 2 MiB; macOS, Windows, WASM and embedded contexts all routinely run threads with smaller budgets, and any of those would hit the overflow on real-world Luau input.

The fix is mechanical: box the heaviest enum variants and the heaviest struct fields throughout the AST so each slot is pointer-sized when its contents aren't being actively constructed. Largest results:

    Stmt:                4784 ->  352 bytes  (~14x)
    Expression:           880 ->  280 bytes  (~3x)
    AnonymousFunction:   2104 ->  168 bytes  (~13x)
    Call:                1144 ->   16 bytes  (~70x)
    TypeAssertion:        872 ->  144 bytes  (~6x)
    ElseIfExpression:     832 ->  288 bytes  (~3x)
    MethodCall:          1144 ->  288 bytes  (~4x)
    FunctionBody:        1944 ->  792 bytes  (~2x)

Stack thresholds for parsing the snippet above:

                  before     after    improvement
    debug         ~2 MiB     ~340 KiB ~6x
    release       ~384 KiB   ~88 KiB  ~4x

For the project's existing deepest stress test (12-level nested do blocks in roblox_cases/pass/multiline_expressions):

                  before     after    improvement
    debug         >2 MiB     ~512 KiB ~4x+
    release       ~256 KiB   ~112 KiB ~2x

Both are now comfortably under any common runtime's thread stack budget.

The remaining stack pressure in debug builds comes from the giant `match` expressions in parse_value and parse_stmt, where rustc reserves frame space for the union of all match arms' locals. That is invariant under AST shape changes and would need a parser refactor (extracting match arms into separate functions) to address, and I want to keep this commit easy to understand and relatively narrow in scope.

Notes:

  - Bumps full-moon to 3.0.0. The variant payload changes are SemVer-major: pattern matches that destructure by value need `Box::new(...)` at construction sites and `*` or auto-deref at binding sites. Code that uses accessor methods or matches by reference is unaffected (Box<T>: Deref<Target = T>).

  - serde representation is unchanged: Box<T> serializes transparently, so all existing snapshot files are byte-identical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [3.0.0] - YYYY-MM-DD
+### Changed
+- **[BREAKING CHANGE]**: shrunk `Stmt`, `Expression`, and other AST nodes by boxing heavy enum variants and struct fields, fixing recursive-parse stack overflows on small-stack threads (#346). `Stmt` is now ~14x smaller (4784 → 352 bytes); `Expression` is ~3x smaller (880 → 280 bytes). The `#346` repro now parses comfortably in a 96 KiB release-mode stack (was ~384 KiB).
+  - Code that destructures variants by value will need `Box::new(...)` at construction sites and `*`/auto-deref at binding sites. Code that uses accessor methods or matches by reference is unaffected, since `Box<T>: Deref<Target = T>`.
+  - `serde` representation is unchanged; `Box<T>` serializes transparently and all existing snapshot files are byte-identical.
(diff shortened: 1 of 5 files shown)


Kampfkarren:
Incredible stuff, thank you so much! Sorry for dropping the ball on reviewing this (and your other PRs), I have had far less time to dedicate to open source these days

PR #356, opened 2026-05-06
Add support for Lua 5.5 global declarations

wez:
Note: this PR is based on top of #355 because I hit issues with stack overflows during testing/implementation.
I'm assuming that #355 will be an easy PR to review and merge and that you might want more discussion on the Lua 5.5 implementation here, which is why that is a separate PR, and I'm assuming that you will ignore the shrink commit when reviewing this PR.

Lua 5.5 introduces a `global` reserved word with two statement forms:

    stat ::= global attnamelist [= explist]
    stat ::= global [attrib] *

It also extends the `attnamelist` rule (shared with `local`)
to allow a *prefix* attribute that applies to every name in the list, in
addition to the existing per-name *postfix* attributes:

    attnamelist ::= [attrib] Name [attrib] {, Name [attrib]}

This commit adds parsing for both, behind a new `lua55` feature flag that
implies `lua54`, and a corresponding `LuaVersion::lua55()` /
`with_lua55()` / `has_lua55()` triple following the existing pattern.

Highlights:

- New `Symbol::Global` reserved word, gated `[lua55]`. Under any
  `LuaVersion` not including 5.5, `global` continues to lex as an
  identifier, preserving backward compatibility for existing 5.1–5.4
  code that uses it as a variable name.

- New `lua55::Global` enum with `GlobalAssignment` / `GlobalWildcard`
  arms, modelling the named and wildcard productions distinctly. Each
  arm carries an optional `prefix_attribute`. `GlobalAssignment` also
  carries per-name `attributes` and (under the `luau` feature)
  `type_specifiers`, mirroring `LocalAssignment`.

- New `Stmt::Global(Box<lua55::Global>)` variant, plus
  `LocalAssignment.prefix_attribute: Option<Box<Attribute>>`.

- The `local` dispatch in `parse_stmt` accepts a `<` next-token under
  5.5 so the prefix-attribute form reaches `expect_local_assignment`.

- `Visit` / `VisitMut` for `GlobalAssignment` are written manually in
  `ast/visitors.rs`, matching the parallel-iteration trick used for
  `LocalAssignment` over name_list / attributes / type_specifiers.

- When both `lua55` and `luau` are active, Luau-style type specifiers
  attach to named globals (e.g. `global x: number = 1`). This is an
  emergent property of full-moon's existing additive-features design,
  not a new dialect — no `lua55+luau` feature flag is introduced.
  Wildcard globals deliberately do not accept type specifiers.

Existing snapshots are unaffected because the new `prefix_attribute`
fields use `skip_serializing_if = "Option::is_none"`. New pass cases
exercise the named, multi-name, prefix-attribute, postfix-attribute,
and wildcard global forms; the `local` prefix-attribute extension; and
the Luau-blended typed-global case. A fail case covers `global = 1`
under 5.5.

Refs: §2.2, §3.1, §3.3.7 of the Lua 5.5 reference manual

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [3.0.0] - YYYY-MM-DD
+### Changed
+- **[BREAKING CHANGE]**: shrunk `Stmt`, `Expression`, and other AST nodes by boxing heavy enum variants and struct fields, fixing recursive-parse stack overflows on small-stack threads (#346). `Stmt` is now ~14x smaller (4784 → 352 bytes); `Expression` is ~3x smaller (880 → 280 bytes). The `#346` repro now parses comfortably in a 96 KiB release-mode stack (was ~384 KiB).
+  - Code that destructures variants by value will need `Box::new(...)` at construction sites and `*`/auto-deref at binding sites. Code that uses accessor methods or matches by reference is unaffected, since `Box<T>: Deref<Target = T>`.
+  - `serde` representation is unchanged; `Box<T>` serializes transparently and all existing snapshot files are byte-identical.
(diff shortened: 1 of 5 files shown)


Kampfkarren:
@wez Looks like just needs rustfmt (ignore clippy)

wez:
Done!

PR #357, opened 2026-05-06
Disambiguate Luau type instantiation from Lua 5.3+ shifts

wez:
When a runtime `LuaVersion` includes both `luau` and `lua53` (or any feature that implies it, like `lua54` or `cfxlua`), the lexer's greedy combining of `<<` and `>>` into single tokens silently breaks every Luau construct that uses adjacent angles:

```
  f<<T>>(x)                                 -- explicit type instantiation
  a:method<<T>>(1)                          -- method type instantiation
  type Foo = Map<string, Array<number>>     -- nested generic close
```

Each of these is two `<` or two `>` tokens in real Luau source. With the lua53 lexer rule active, they're combined into `DoubleLessThan` / `DoubleGreaterThan`, the parser doesn't recognise the combined token in suffix position, and the user gets a confusing parse error.

This is an emergent ambiguity from full-moon's additive feature flags. Lua 5.3+ has shifts but no generics; Luau has generics but no shifts. No real source file has both. The blend exists because some callers want to parse Luau-flavoured code with Lua 5.4 attributes (or other non-Luau extensions), and have to enable both feature sets to do so.

Fix: when `has_luau()` is true on the runtime `LuaVersion`, suppress the lua53/cfxlua `<<` and `>>` combining in the lexer. Adjacent angles emerge as separate tokens, which the existing type-instantiation and generic-type-params parsers already handle correctly. Bitwise shifts and compound shift-assigns (`<<=`, `>>=`) become unavailable under that exact feature combination as a trade-off — there's no realistic source where both are needed.

Behavior summary:

```
  feature combination    `<<`/`>>` parses as
  --------------------   ----------------------
  lua53 (or 54, cfxlua)  shift / compound-assign      (unchanged)
  luau only              two single-angle tokens      (unchanged)
  luau + lua53/54/cfx    two single-angle tokens      (was: parse error)
```

Pure Lua 5.3+ and pure Luau parsing are byte-identical to before. The change only affects the blended runtime configuration.

Regression tests added in `tests/luau_shift_disambiguation.rs` covering type instantiation, method type instantiation, nested generics, and triple-nested generics, plus a negative test confirming shifts still parse under `lua53` alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Fixed
+- When a runtime `LuaVersion` has both Luau and Lua 5.3+ enabled, the lexer no longer combines `<<` and `>>` into single tokens. This lets blended-dialect callers parse Luau's explicit type-instantiation syntax (`f<<T>>(x)`) and nested generic types (`Map<K, Vec<V>>`) correctly. Bitwise shifts and compound shift-assigns (`<<=`, `>>=`) become unavailable under that exact feature combination as a trade-off; pure Lua 5.3/5.4 and pure Luau parsing are unaffected.
 
 ## [2.2.0] - 2026-04-15
 ### Added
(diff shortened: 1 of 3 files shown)


Kampfkarren:
FWIW the blended feature stuff is mostly just to do our best effort at parsing any .lua file that comes our way. It's not expected anyone's going to actually use multiple languages together.

wez:
I made a thing that does blend a couple of these together: https://shingetsu.dev/differences/ which is what had me poking around at this!

PR #358, opened 2026-05-10
luau: allow separators before number prefixes and in exponents

bytexenon:
Luau allows underscores between the leading zero and the number prefix:

```luau
return 0_x40 --> 64
return 0__b11 --> 3
```

The lexer already supports separators in numbers, but only *after* the prefix. This PR adds support for separators *before* the prefix as well.

### DIFF
--- a/full-moon/src/tokenizer/lexer.rs
+++ b/full-moon/src/tokenizer/lexer.rs
@@ -335,16 +335,28 @@ impl Lexer {
             }
 
             initial @ '0' => {
+                let mut number = String::from(initial);
+                if self.lua_version.has_luau() {
+                    while let Some(cur) = self.source.current() {
+                        if cur == '_' {
+                            number.push(cur);
+                            self.source.next();
+                        } else {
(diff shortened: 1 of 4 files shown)


bytexenon:
Actually, I just noticed the lexer is also missing support for underscores after the exponent part (e.g., `1e_2` or `1e+_2` for 100). I'm converting this PR to a draft temporarily so I can add that fix in here as well.

bytexenon:
This should be complete now

I also added Luau coverage to the existing decimal_seperators pass case for the extra forms discussed above:

```patch
local num1 = 1_048_576
local num2 = 0xFFFF_FFFF
local num3 = 0b_0101_0101
local num4 = 1_523_423.132_452_312
local num5 = 1e512_412
local num6 = 1e-512_412
+ local num7 = 0_xFF_FF_FF_FF
+ local num8 = 0_b_1010_1010_1010_1010
+ local num9 = 1e+_2
```

